PR #2, opened 2013-05-10
support modern libqhull libraries

v4hn:
The module was renamed from 'qhull' to 'libqhull' and therefore
geometric_shapes fails to compile on up2date systems.

This commit adds a proper cmake test module for various qhull versions (taken
from pcl) and passes the detected version on via a config file. That way the
resp. correct headers are included.


isucan:
thank you!

PR #3, opened 2013-05-16
Add function to compute bounding sphere of a shape

acornacorn:


src/shape_operations.cpp
```diff
@@ -301,6 +301,70 @@ class ShapeVisitorComputeExtents : public boost::static_visitor<Eigen::Vector3d>
...
+  radius = 0.0;
+
+  if (shape->type == SPHERE)
+  {
+    radius = static_cast<const Sphere*>(shape)->radius;
+  }
+  else if (shape->type == BOX)
+  { 
+    const double* sz = static_cast<const Box*>(shape)->size;
+    radius = std::sqrt(sz[0] * sz[0] + sz[1] * sz[1] + sz[2] * sz[2]);
```
isucan: should this not be divided by 2? size here is the full length of a box side

src/shape_operations.cpp
```diff
@@ -301,6 +301,70 @@ class ShapeVisitorComputeExtents : public boost::static_visitor<Eigen::Vector3d>
...
+  }
+  else if (shape->type == CYLINDER)
+  {
+    double r = static_cast<const Cylinder*>(shape)->radius;
+    double l = static_cast<const Cylinder*>(shape)->length;
+    radius = std::sqrt(r * r + l * l);
```
isucan: I may be wrong here, but I think this should be something like 1/2 distance from 0 0 0 to 2r, 2r, l (thinking of a box containing the cylinder)
and this would be sqrt(4r^2 + 4r^2 + l^2)/2 = sqrt(8r^2+l^2)/2

src/shape_operations.cpp
```diff
@@ -301,6 +301,70 @@ class ShapeVisitorComputeExtents : public boost::static_visitor<Eigen::Vector3d>
...
+  }
+  else if (shape->type == CONE)
+  {
+    double r = static_cast<const Cone*>(shape)->radius;
+    double l = static_cast<const Cone*>(shape)->length;
+
+    if (l > r)
```
isucan: i did not check this

src/shape_operations.cpp
acornacorn: A sphere circumscribing a box that circumscribes a cylinder is much bigger than the sphere that circumscribes the cylinder.  What we want is the sphere that circumscribes the cylinder which is (I believe) what I have here.  The sphere will be centered at the center of the cylinder, and the circle at each end of the cylinder will be on the surface of the sphere.

src/shape_operations.cpp
acornacorn: I'll fix...

src/shape_operations.cpp
isucan: I think if we try to be tight, the answer is 
sqrt(r^2 +l^2/4)

(l is the total height of the cylinder)

acornacorn: yes, you are right.  Will fix.

src/shape_operations.cpp
```diff
@@ -301,6 +301,73 @@ class ShapeVisitorComputeExtents : public boost::static_visitor<Eigen::Vector3d>
...
+    {
+      // center of sphere is intersection of perpendicular bisectors:
+      double z = (l - (r * r / l)) * 0.5;
+      center.z() = z - (l * 0.5);
+      radius = l - z;
+    }
+    else
+    {
+      // short cone.  Bounding sphere touches base only.
+      center.z() = - (l * 0.5);
```
isucan: should this not be 0? If the radius of the sphere is the same as the base of the cone, the ecuator of the sphere must be the same as the base circle of the cone, right?

src/shape_operations.cpp
```diff
@@ -301,6 +301,73 @@ class ShapeVisitorComputeExtents : public boost::static_visitor<Eigen::Vector3d>
...
+    radius = std::sqrt(r * r + l * l);
+  }
+  else if (shape->type == CONE)
+  {
+    double r = static_cast<const Cone*>(shape)->radius;
+    double l = static_cast<const Cone*>(shape)->length;
+
+    if (l > r)
+    {
+      // center of sphere is intersection of perpendicular bisectors:
```
isucan, line 339: Should this case maybe be handled as the cylinder? Can we do any better than that?

src/shape_operations.cpp
```diff
@@ -301,6 +301,73 @@ class ShapeVisitorComputeExtents : public boost::static_visitor<Eigen::Vector3d>
...
+    double x = sz[0] * 0.5;
+    double y = sz[1] * 0.5;
+    double z = sz[2] * 0.5;
+    radius = std::sqrt(x*x + y*y + z*z);
+  }
+  else if (shape->type == CYLINDER)
+  {
+    double r = static_cast<const Cylinder*>(shape)->radius;
+    double lo2 = static_cast<const Cylinder*>(shape)->length * 0.5;
+    radius = std::sqrt(r * r + l * l);
```
isucan: I am looking at the diff only, but does this compile? It seems you shold have lo2 instead of l

src/shape_operations.cpp
acornacorn: I think the center of the shape is halfway between the tip of the cone and the center of the cone's base circle.  Therefore the center of the circle is at z=-len/2 right?  So I think this is correct.

src/shape_operations.cpp
acornacorn, line 339: We can do better than a cylinder.  In this implementation the circular base of the cone is on the sphere and the tip of the cone is also on the sphere.  I believe that is the smallest circle that contains the cone.

src/shape_operations.cpp
acornacorn: I am writing some unit tests.  At some point I should try this in rviz and verify this visially, but I have not gotten around to that yet.

PR #20, opened 2014-06-25
update usage of console_bridge to deal with version in Trusty

wjwwood:
This is required because of https://github.com/ros/rosdistro/issues/4633

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,8 @@ else()
 endif()
 
 find_package(octomap REQUIRED)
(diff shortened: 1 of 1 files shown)


wjwwood:
@dirk-thomas @tfoote 

Also... Why am I not in the ros-planning group?

tfoote:
I have merged this into an indigo-devel branch on top of the latest commits and released as 0.4.0. I did not change the dependencies so it could be backwards compatible. Just doing this as an emergency patch to get indigo building again.

PR #21, opened 2014-06-26
update to use debian console_bridge dependency.

tfoote:
https://github.com/ros/rosdistro/issues/4633


dirk-thomas:
Btw. travis is failing since the _indigo-devel_ branch is testing against _precise_ - instead it should test on _trusty_ (https://github.com/ros-planning/geometric_shapes/blob/14d2ab4e341bf7d92d7516b8deb0f5700bace703/.travis.yml#L23).

PR #30, opened 2015-03-27
Fix FindQhull with non-debian systems

v4hn:
See https://github.com/PointCloudLibrary/pcl/pull/852 for details on this bug.


v4hn:
guys? @mikeferguson, whoever is responsible?
Could you please consider&merge/discuss this request?

v4hn:
Ok, I just checked out a clean copy of geometric_shapes, tried to run move_group with it, and got a segfault.
After 5 minutes I realized it's the very same issue I spent hours on debugging last time.
As I'm rather sure I'm the only developer who cares about non-Ubuntu support, and as I just made sure this does not break with ubuntu 16.04. and 14.04. I'll merge it now.

PR #31, opened 2015-04-09
Add dependency

jspricke:


mikeferguson:
For completeness, this should also probably add it to the CMakeLists find_package(catkin) call, no?

jspricke:
Oups, must have been to late yesterday ;). Should be fixed.

PR #35, opened 2015-11-22
Use catkin variables for install destination

jspricke:


v4hn:
Thanks for spotting this.
Obviously catkin's variables should be used for install destinations.

PR #41, opened 2016-06-15
append cmake module path instead of prepending

### LINKED ISSUE #22: Appending to CMAKE_MODULE_PATH
I would like to specify my own FindQhull.cmake to your package externally, as I have the qhull include files and libraries in special paths (due to package management). This is currently not possible as you prepend your custom cmake module path to CMAKE_MODULE_PATH in https://github.com/ros-planning/geometric_shapes/blob/hydro-devel/CMakeLists.txt#L4. Would it be possible to instead append to the CMAKE_MODULE_PATH? This will only change the cmake find behaviour if someone else specifes a FindQhull.cmake file on the path as the CMAKE_MODULE_PATH is empty by default. If yes, I can provide a pull request.

v4hn:
Fixes #22

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(geometric_shapes)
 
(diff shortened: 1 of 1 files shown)


v4hn:
superseded by #43

PR #50, opened 2016-08-02
Export C++11 compiler flag

jonbinney:
I'd recommend waiting on merging this until we get a bit more feedback on https://github.com/ros-planning/geometric_shapes/issues/49

If we decide to keep `std::shared_ptr` in the API of geometric_shapes, this PR exports the `-std=c++11` compiler flag so that packages that depend on geometric_shapes will compile properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,8 @@ catkin_package(
     EIGEN3
     OCTOMAP
     console_bridge
+  CFG_EXTRAS
+    geometric_shapes-extras.cmake
   )
 
 find_package(Qhull REQUIRED)
--- a/cmake/geometric_shapes-extras.cmake
+++ b/cmake/geometric_shapes-extras.cmake
@@ -1,2 +1,2 @@
-set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
+add_compile_options("-std=c++11")
 
--- a/cmake/geometric_shapes-extras.cmake
+++ b/cmake/geometric_shapes-extras.cmake
@@ -1,0 +1,2 @@
+set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
+


v4hn:
Interesting way to expose the flag. where did you find that? :)
Is the -extras code implicitly executed?

Is there a reason why this doesn't use `add_compile_options` as discussed in https://github.com/ros-planning/moveit_planners/pull/73 ?

jonbinney:
I followed the suggested solution here: http://answers.ros.org/question/173947/export-gcc-flags-though-catkin-cfg_extras/ The `-extras` code gets run any time someone `find_packages` geometric_shapes.

I don't actually have a good reason for modifying CMAKE_CXX_FLAGS instead of using add_compile_options, other than it was how I saw it done in @dirk-thomas's ROS answers post. I'm happy to change it if add_compile_options is more correct.

v4hn:
> I'm happy to change it if add_compile_options is more correct.

Do that please.

In 2014 it was not necessarily ROS-safe to use that command yet, because the minimum cmake version in hydro was 2.8.3.

jonbinney:
@v4hn I've added a commit to this PR that switches it to `add_compile_options`

de-vri-es:
I'm not sure that exporting the flag to packages depending on `geometric_shapes` is really a good idea. It makes it more difficult for people to compile with a different (compatible) standard such as `gnu++11`, `c++14`, `gnu++14` (which is the default in gcc 6) etc.

I understand that exporting the flag this way saves adding it in every other package using it, but I also feel it should be a conscious decision made by package maintainers rather than something sneaked in by a dependency. Especially considering that there are valid alternatives to `-std=c++11`.

v4hn:
@de-vri-es we _do_ require C++11 in our public headers.
So either _we_ decide to export a standard or everyone who uses the geometric_shapes module will have to explicitly decide and add it manually to their compile flags.
As they might not need it in their own code, I don't think it should be in their build scripts.

Importantly, everyone can simply override the flag by adding a different standard via `add_compile_options`. With gcc the last specified standard wins and is used.

de-vri-es:
I understand the reasoning, and it is true that you can overwrite the standard used after invoking `find_package(catkin REQUIRED COMPONENTS geometric_shapes)`. But it does mean that you now have to know which package exports a standard and fix it afterwards if desired.

For example, at work we have one package to set a bunch of compiler options (`-Wall -Wextra` and nowadays `-std=c++14`). So with this change we need would need to list the package in a specific order to have the desired options win.

If you're aware of the fact that `geometric_shapes` exports `-std=c++11`, this is only a nuisance (though not a minor one in my opinion, I like my dependencies sorted by alphabet ^_^). But if you're not aware of `geometric_shapes` exporting `std=c++11`, this is annoying to debug and possibly long and frustrating search for the solution. (Granted, an experienced person should spot the problem relatively quickly by inspecting the compiler commands and then pinpointing the problem to `geometric_shapes`, but not everyone is equally experienced.) 

On the other hand, if the option is not exported, the compiler gives a reasonably clear error message that should quickly point to compiling with C++11 support.

jonbinney:
@de-vri-es i think I agree with you on this.  I'm leaving this PR open as one example of how to "fix" the compile problems, but there are at least two other possible solutions:
- require packages that depend on geometric_shapes to add the `-std=c++11` flag themselves (as you point out). This has the disadvantage of needing to modify all package that depend on geometric_shapes, but the advantage that it is more "explicit"
- find some way to avoid c++11 specific features in the geometric_shapes API and headers. This has the advantage of following REP 0003, but has some challenges as discussed here: https://github.com/ros-planning/geometric_shapes/issues/49

In any case this is a pretty important decision, and based on https://github.com/ros-planning/geometric_shapes/issues/49 it sounds like the next step is to have a discussion on the ROS Discourse to choose the right high-level solution. I'd suggest leaving this PR here for now as an example possible solution that people can point to.

v4hn:
On Mon, Aug 08, 2016 at 05:58:00PM -0700, Jonathan Binney wrote:

> it sounds like the next step is to have a discussion on the ROS Discourse to choose the right high-level solution.

Feel free to start a thread there.

de-vri-es:
See http://discourse.ros.org/t/std-shared-ptr-in-the-geometric-shapes-public-api/395

mryellow:
Note that with https://github.com/ros-planning/geometric_shapes/pull/50 merged `collada_urdf` will build successfully if edited to use `__typeof__` instead of `typeof` .

v4hn:
@mryellow thanks for testing, but by now we pretty much agreed in http://discourse.ros.org/t/std-shared-ptr-in-the-geometric-shapes-public-api/395/10 that we will not export the flag here.
Instead, the depending projects have to compile with C++11 or newer.

Closing as no-merge.

PR #52, opened 2016-08-25
Clear root transformation on imported Collada meshes.

de-vri-es:
Assimp enforces a Y_UP convention when importing file formats that store this metadata, such as the collada format with the `<up_axis>` tag. If an imported model is using a different convention such as Z_UP, the whole model is rotated until the Y axis of the model is pointing up [1]. This is both confusing and not in-line with the ROS convention where the Z axis is pointing up.

Even worse, it causes collision checks on most collada models to use this wrongly rotated model. Rviz is already undoing this rotation[2], so the collision mesh used by moveit doesn't even match the collision mesh as visualized by rviz.

It is conceivable though that the root node transformation may be used for legit purposes with other file types, so this PR limits itself to files with `dae` or `zae` extension.

I made a small test case which publishes three meshes as markers [3]: a collada model with Z_UP, one with Y_UP and an stl model. Without this patch the Z_UP model is rotated, with the patch all three models are the same and the metadata is basically ignored.

Another option would be to assume Y is pointing up after import and rotate the model to make Z point up again. This would be in line with the ROS convention. However, this would still leave a mismatch with rviz, and frankly I think that Assimp adjusting the model orientation based on this metadata is misguided in the first place. I think that what was X in the CAD program or 3D modelling software should remain X after importing.

[1] https://github.com/assimp/assimp/blob/23baecaff30b7a280648834368b1c0c792e4092e/code/ColladaLoader.cpp#L192
[2] https://github.com/ros-visualization/rviz/blob/0d423d6249a39f7f3ddd591007952c6dc940b6ac/src/rviz/mesh_loader.cpp#L232
[3] https://github.com/de-vri-es/test_load_mesh

### DIFF
--- a/src/mesh_operations.cpp
+++ b/src/mesh_operations.cpp
@@ -243,14 +243,23 @@ Mesh* createMeshFromBinary(const char *buffer, std::size_t size, const Eigen::Ve
   const aiScene* scene = importer.ReadFileFromMemory(reinterpret_cast<const void*>(buffer), size,
                                                      aiProcess_Triangulate            |
                                                      aiProcess_JoinIdenticalVertices  |
-                                                     aiProcess_SortByPType            |
-                                                     aiProcess_OptimizeGraph          |
-                                                     aiProcess_OptimizeMeshes,
+                                                     aiProcess_SortByPType,
                                                      hint.c_str());
-  if (scene)
-    return createMeshFromAsset(scene, scale, hint);
-  else
+  if (!scene)
     return NULL;
+
+  // Assimp enforces Y_UP convention by rotating models with different conventions.
+  // However, that behaviour is confusing and doesn't match the ROS convention
+  // where the Z axis is pointing up.
+  // Hopefully this doesn't undo legit use of the root node transformation...
+  // Note that this is also what RViz does internally.
+  scene->mRootNode->mTransformation = aiMatrix4x4();
+
+  // These post processing steps flatten the root node transformation into child nodes,
+  // so they must be delayed until after clearing the root node transform above.
+  importer.ApplyPostProcessing(aiProcess_OptimizeMeshes | aiProcess_OptimizeGraph);
+
+  return createMeshFromAsset(scene, scale, hint);
 }
 
 Mesh* createMeshFromResource(const std::string& resource, const Eigen::Vector3d &scale)
--- a/test/test_create_mesh.cpp
+++ b/test/test_create_mesh.cpp
@@ -1,0 +1,102 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2016, Delft Robotics
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the copyright holder nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+/** \Author Maarten de Vries <maarten@de-vri.es> */
+
+#include "resources/config.h"
+#include <geometric_shapes/mesh_operations.h>
+#include <gtest/gtest.h>
+#include <string>
+
+namespace {
+
+  void assertMesh(shapes::Mesh * mesh)
+  {
+    ASSERT_TRUE(mesh != NULL);
+    ASSERT_EQ(3, mesh->vertex_count);
+    ASSERT_EQ(1, mesh->triangle_count);
+
+    ASSERT_EQ(0, mesh->triangles[0]);
+    ASSERT_EQ(1, mesh->triangles[1]);
+    ASSERT_EQ(2, mesh->triangles[2]);
+
+    ASSERT_FLOAT_EQ(0, mesh->vertices[0 + 0]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[0 + 1]);
+    ASSERT_FLOAT_EQ(1, mesh->vertices[0 + 2]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[3 + 0]);
+    ASSERT_FLOAT_EQ(1, mesh->vertices[3 + 1]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[3 + 2]);
+    ASSERT_FLOAT_EQ(1, mesh->vertices[6 + 0]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[6 + 1]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[6 + 2]);
+  }
+
+  shapes::Mesh * loadMesh(const std::string & mesh)
+  {
+    std::string path = "file://" + std::string(TEST_RESOURCES_DIR) + "/" + mesh;
+    return shapes::createMeshFromResource(path);
+  }
+
+}
+
+TEST(CreateMesh, stl)
+{
+  assertMesh(loadMesh("triangle.stl"));
+}
+
+TEST(CreateMesh, daeNoUp)
+{
+  assertMesh(loadMesh("triangle_no_up.dae"));
+}
+
+TEST(CreateMesh, daeYUp)
+{
+  assertMesh(loadMesh("triangle_y_up.dae"));
+}
+
+TEST(CreateMesh, daeZUp)
+{
+  assertMesh(loadMesh("triangle_z_up.dae"));
+}
+
+TEST(CreateMesh, daeXUp)
+{
+  assertMesh(loadMesh("triangle_x_up.dae"));
+}
+
+int main(int argc, char **argv)
+{
+  testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
--- a/test/test_create_mesh.cpp
+++ b/test/test_create_mesh.cpp
@@ -95,6 +95,21 @@ TEST(CreateMesh, daeXUp)
   assertMesh(loadMesh("triangle_x_up.dae"));
 }
 
+TEST(CreateMesh, daeNoUnit)
+{
+  assertMesh(loadMesh("triangle_no_unit.dae"));
+}
+
+TEST(CreateMesh, dae1M)
+{
+  assertMesh(loadMesh("triangle_1m.dae"));
+}
+
+TEST(CreateMesh, dae10M)
+{
+  assertMesh(loadMesh("triangle_10m.dae"));
+}
+
 int main(int argc, char **argv)
 {
   testing::InitGoogleTest(&argc, argv);
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -11,3 +11,6 @@ target_link_libraries(test_point_inclusion ${PROJECT_NAME} ${catkin_LIBRARIES} $
 
 catkin_add_gtest(test_bounding_sphere test_bounding_sphere.cpp)
 target_link_libraries(test_bounding_sphere ${PROJECT_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+
+catkin_add_gtest(test_create_mesh test_create_mesh.cpp)
+target_link_libraries(test_create_mesh ${PROJECT_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
--- a/test/resources/triangle.stl
+++ b/test/resources/triangle.stl
@@ -1,0 +1,9 @@
+solid
+facet normal -0.577350 -0.577350 -0.577350
+outer loop
+vertex 0.000000 0.000000 1.000000
+vertex 0.000000 1.000000 0.000000
+vertex 1.000000 0.000000 0.000000
+endloop
+endfacet
+endsolid
--- a/test/resources/triangle_10m.dae
+++ b/test/resources/triangle_10m.dae
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
+<COLLADA xmlns="http://www.collada.org/2005/11/COLLADASchema" version="1.4.1">
+  <asset>
+    <created>2016-08-28T22:17:50</created>
+    <modified>2016-08-28T22:17:50</modified>
+    <unit name="decameter" meter="10"/>
+  </asset>
+  <library_images/>
+  <library_geometries>
+    <geometry id="triangle">
+      <mesh>
+        <source id="positions">
+          <float_array id="data" count="9">0 0 1 0 1 0 1 0 0</float_array>
+          <technique_common>
+            <accessor source="#data" count="3" stride="3">
+              <param name="X" type="float"/>
+              <param name="Y" type="float"/>
+              <param name="Z" type="float"/>
+            </accessor>
+          </technique_common>
+        </source>
+        <vertices id="vertices"><input semantic="POSITION" source="#positions"/></vertices>
+        <triangles count="1">
+          <input semantic="VERTEX" source="#vertices" offset="0"/>
+          <p>0 1 2</p>
+        </triangles>
+      </mesh>
+    </geometry>
+  </library_geometries>
+  <library_controllers/>
+  <library_visual_scenes>
+    <visual_scene id="scene"><node><instance_geometry url="#triangle"/></node></visual_scene>
+  </library_visual_scenes>
+  <scene><instance_visual_scene url="#scene"/></scene>
+</COLLADA>
--- a/test/resources/triangle_1m.dae
+++ b/test/resources/triangle_1m.dae
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
+<COLLADA xmlns="http://www.collada.org/2005/11/COLLADASchema" version="1.4.1">
+  <asset>
+    <created>2016-08-28T22:17:50</created>
+    <modified>2016-08-28T22:17:50</modified>
+    <unit name="meter" meter="1"/>
+  </asset>
+  <library_images/>
+  <library_geometries>
+    <geometry id="triangle">
+      <mesh>
+        <source id="positions">
+          <float_array id="data" count="9">0 0 1 0 1 0 1 0 0</float_array>
+          <technique_common>
+            <accessor source="#data" count="3" stride="3">
+              <param name="X" type="float"/>
+              <param name="Y" type="float"/>
+              <param name="Z" type="float"/>
+            </accessor>
+          </technique_common>
+        </source>
+        <vertices id="vertices"><input semantic="POSITION" source="#positions"/></vertices>
+        <triangles count="1">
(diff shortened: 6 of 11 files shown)


de-vri-es:
Not confirmed yet, but I think this moveit issue would be solved by this PR: ros-planning/moveit#159

v4hn:
Wow, dae loading is badly fubar...

In my opinion our code has to load _all_ models the same way RViz does.
People use RViz for the validation of their URDFs, so this is what they expect.
If this request achieves this +1 to the idea, I don't know enough about asset loading
to review the request though.

de-vri-es:
Fair enough, I agree that what rviz shows should be what moveit and others use internally as well. That basically means removing the `if (hint == ...)`.

It also means tracking any changes made by rviz in the future, although it is not likely that rviz will change this in the near future. If they do we'll have a pretty nasty backwards compatibility problem.

de-vri-es:
Updated to unconditionally clear the root node transformation, same as rviz.

v4hn:
On Fri, Aug 26, 2016 at 03:37:58AM -0700, Maarten de Vries wrote:

> Updated to unconditionally clear the root node transformation, same as rviz.

Alternatively you are surely very welcome to fix this in RViz kinetic and then update it here. :-)

de-vri-es:
I'm afraid that what rviz does is probably the best solution. As far as I can tell the root node transformation is in practise only used for things like "fixing" the orientation and "correcting" for unit scale. However, all of those adjustments mean that 1 unit along the X, Y or Z axes of the 3D modelling software is not the same as 1 unit along the same axes after importing. I would say that undoing these "corrections" is best.

The only problem is that there are no guarantees that the root node transformation is only used for these metadata adjustments. I don't know if there are importers now that use the root node transformation for something we'd want to keep, let alone what importers may do in the future.

Still, the current situation has to be fixed, and I don't see a better solution than simply clearing the root node transformation.

A long term solution should be implemented in Assimp, possibly in the form of a flag to disable the use of this type of metadata. I can poke around there, but at best it will be a while before that is universally available.

v4hn:
Again, I can't review this properly, but could you please add a comment saying that this behavior is supposes to match RViz?

de-vri-es:
Comment added.

Also, because a picture is worth a thousand words:

Unpatched:
collada_unpatched_geometric_shapes

Patched:
collada_patched_geometric_shapes

Both images generated with the test case at [1]. The first one without this PR, the second with. The three shapes are all the same model, the first two collada with Z_UP and Y_UP respectively, the third one the same model as STL.

[1] https://github.com/de-vri-es/test_load_mesh

davetcoleman:
This could potentially break applications that have worked around this bug by a hacked transform flip, correct? So we are only targeting this for kinetic and maybe jade?

Perhaps @wjwwood, maintainer of Rviz, could weight in his opinion.

I love the pictures!

You mentioned a small test case - think you could turn that into a unit test in this PR?

Thanks for looking into this!

src/mesh_operations.cpp
```diff
@@ -243,14 +243,23 @@ Mesh* createMeshFromBinary(const char *buffer, std::size_t size, const Eigen::Ve
   const aiScene* scene = importer.ReadFileFromMemory(reinterpret_cast<const void*>(buffer), size,
                                                      aiProcess_Triangulate            |
                                                      aiProcess_JoinIdenticalVertices  |
-                                                     aiProcess_SortByPType            |
-                                                     aiProcess_OptimizeGraph          |
-                                                     aiProcess_OptimizeMeshes,
```
davetcoleman, line 248: why remove these optimizations? is this the part that rotates the axis?

de-vri-es: They're not removed, just moved down later. Those two optimizations heavily optimize the imported meshes and scene graph. One of the results was that the root node transformation got flattened into child nodes. That made clearing the root node transformation a NOP, but clearing the child node transformations is not an option since it would modify the mesh.

But the optimizations are still performed after clearing the root node transform here: https://github.com/ros-planning/geometric_shapes/pull/52/files#diff-f5ec0f576aaf3b44a4c42c739e51980dR260

de-vri-es:
> This could potentially break applications that have worked around this bug by a hacked transform flip, correct? So we are only targeting this for kinetic and maybe jade?

I don't think any program could have worked around it reliably after the mesh left this function. You need knowledge about the loaded file to undo the effect: best would be the root node transformation or otherwise you should know if the imported file was a collada model with something other than `<up_axis>Y_UP</up_axis>`.

That does leave the possibility that someone worked around it _unreliably_, although I think most people would have just rotated the mesh and/or export as STL rather than apply a rotation in code. Still, it is worth considering. In the strictest sense it is indeed a backward incompatible change.

> You mentioned a small test case - think you could turn that into a unit test in this PR?

+1. The test case so far was geared for visual inspection, but it is possible to make a unit test for this with simpler meshes. Considering the importance of correctly importing meshes for collision models and the possibility that Assimp might change the behaviour in the future, I agree that a unit test is a very good idea. I'll start work on one  :)

de-vri-es:
> That does leave the possibility that someone worked around it unreliably, although I think most people would have just rotated the mesh and/or export as STL rather than apply a rotation in code. Still, it is worth considering. In the strictest sense it is indeed a backward incompatible change.

I just realised that the workaround of rotating the collada mesh will still be broken by this change. They would have had to notice the wrong orientation though, which still wasn't very easy considering RViz showed the expected collision model.

mikeferguson:
@de-vri-es Was there previously any way to actually visualize the mesh when using a workaround? Presumably all the RVIZ plugins showed the incorrect rotation, right? My expectation is that people have probably fixed their meshes to work around the bug?

de-vri-es:
> @de-vri-es Was there previously any way to actually visualize the mesh when using a workaround? Presumably all the RVIZ plugins showed the incorrect rotation, right? My expectation is that people have probably fixed their meshes to work around the bug?

I believe the RViz plugins actually show the correct rotation. They don't seem to be using the mesh as imported by `geometric_shapes`. To get the wrong rotation visible, the mesh would have to have been converted to a marker and send to RViz, as my test case does.

/update: Although I must add that I haven't used the rviz plugins in a while.

de-vri-es:
To clarify: the moveit rviz plugins seem to use `rviz::Robot` as a base for visualizing robots, which uses RViz's method of loading meshes.

v4hn:
On Fri, Aug 26, 2016 at 11:12:15AM -0700, Dave Coleman wrote:

> This could potentially break applications that have worked around this bug by a hacked transform flip, correct?
> So we are only targeting this for kinetic and maybe jade?

This is clearly a bug fix and probably even @hersh tripped over this without understanding the problem.
So if someone analyzed this, _and_ worked around it _and_ never even cared to report it, it's definitely their very own fault.

Bug Fixes should _never_ not be merged because someone might have worked around the broken behavior.

src/mesh_operations.cpp
davetcoleman: got it

davetcoleman:
> That does leave the possibility that someone worked around it unreliably

I probably have done this in the past, but its no problem to undo the hack

> Bug Fixes should _never_ not be merged because someone might have worked around the broken behavior.

I wouldn't say never, and I think there's lots of examples of the contrary, but in this case I agree

mikeferguson:
My suggestion would be to make sure we announce it on the moveit mailing list that we have fixed/merged this -- and since I/J syncs just happened yesterday, we will have significant soak time for anyone to test, catch, and fix regressions that might occur with their robots before it goes live to the world.

v4hn:
> I wouldn't say never, and I think there's lots of examples of the contrary, but in this case I agree

Citing the by now classic [Ulrich Drepper's "Good Practices in Library Design"](http://www.akkadia.org/drepper/goodpractice.pdf):

> One sort of changes which should be made without hesitation are bug fixes. They
> should be made even if the API is changed. Some people argue that some programs are
> depending on the old, broken behavior. But this argumentation is flawed. Equally well
> there are programs which depend on the correct behavior, written by people who read
> the specification. No program which depends on broken behavior deserves protection.
> If a programmer has found such a discrepancy and changed the application to use it
> instead of reporting it to the library maintainer s/he gets what s/he deserves.

I fully agree with him.

> My suggestion would be to make sure we announce it on the moveit mailing list

Definitely!

test/test_create_mesh.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+/** \Author Maarten de Vries <maarten@de-vri.es> */
+
+#include <geometric_shapes/mesh_operations.h>
+#include <gtest/gtest.h>
+#include <string>
+
+namespace {
+
+std::string makeDaeModel(std::string const & up_axis) {
```
de-vri-es: Is this an accepted method to embed text files into tests, or should I put it in a file loaded at runtime?

I like how this is self-contained without any runtime dependencies on the filesystem or environment, but at the same time it isn't the prettiest in code.

/edit: Best to look at the full changes, otherwise you only see the function signature.

de-vri-es:
I added a unit test to verify loading a simple mesh (see also the line comment regarding embedding the mesh).

One more important thing to note about this PR: it also undoes scaling from the `<unit>` metadata tag. Again, this is the same thing RViz does, and I think this behaviour matches expectations best, but it is an important detail nonetheless.

For completeness the unit test should also test STL, and verify that the `<unit>` tag is indeed ignored. I will add those later, but I would like feedback on the line comment :)

test/test_create_mesh.cpp
v4hn: On Sun, Aug 28, 2016 at 01:48:27PM -0700, Maarten de Vries wrote:

> Is this an accepted method to embed text files into tests, or should I put it in a file loaded at runtime?

This makes it quite hard to load the same dae file with a different tool, right..
I would prefer a solution where the test files are separate and complete (i.e. with a specified direction).

de-vri-es: True. What is the preferred method of making the file available at runtime then? Use a `package://` URL (already supported by `geometric_shapes`, so no new dependency) and require the environment to be set up with a proper `ROS_PACKAGE_PATH`? I imagine that the `run_tests` target of the generated `Makefile` already sets up the environment correctly?

Alternatively the file can exist in the source tree and be linked in to the executable as a binary blob, but I imagine this is less standard for ROS packages.

v4hn: I believe @rhaschke recently worked with external files & unit tests. Could you please comment?

de-vri-es:
Updated tests to use files and the already available `#include "resources/config.h"`. Also added an STL version of the mesh.

test/test_create_mesh.cpp
davetcoleman: @rhaschke used some really fancy CMake magic:
https://github.com/ros-planning/moveit_resources/blob/master/CMakeLists.txt
https://github.com/ros-planning/moveit/commit/b537062ad9f480701b89a6f83cceb4a4159462e0

But the simpler method works fine too:
https://github.com/ros-planning/moveit/blob/c8947005264580e038510b5d9460ec2226313153/moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp#L64

test/test_create_mesh.cpp
```diff
@@ -0,0 +1,100 @@
...
+/** \Author Maarten de Vries <maarten@de-vri.es> */
+
+#include "resources/config.h"
+#include <geometric_shapes/mesh_operations.h>
+#include <gtest/gtest.h>
+#include <string>
+
+namespace {
+
+  void assertMesh(shapes::Mesh * mesh) {
```
davetcoleman: bracket on new line

test/test_create_mesh.cpp
```diff
@@ -0,0 +1,100 @@
...
+    ASSERT_FLOAT_EQ(1, mesh->vertices[3 + 1]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[3 + 2]);
+    ASSERT_FLOAT_EQ(1, mesh->vertices[6 + 0]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[6 + 1]);
+    ASSERT_FLOAT_EQ(0, mesh->vertices[6 + 2]);
+  }
+
+  shapes::Mesh * loadMesh(const std::string & path)
+  {
+      return shapes::createMeshFromResource("file://" + std::string(TEST_RESOURCES_DIR) + "/" + path);
```
davetcoleman: indent of 4 spaces...

davetcoleman: this method for loading meshes looks perfect

davetcoleman:
```
[==========] Running 5 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 5 tests from CreateMesh
[ RUN      ] CreateMesh.stl
[       OK ] CreateMesh.stl (1 ms)
[ RUN      ] CreateMesh.daeNoUp
[       OK ] CreateMesh.daeNoUp (0 ms)
[ RUN      ] CreateMesh.daeYUp
[       OK ] CreateMesh.daeYUp (0 ms)
[ RUN      ] CreateMesh.daeZUp
[       OK ] CreateMesh.daeZUp (1 ms)
[ RUN      ] CreateMesh.daeXUp
[       OK ] CreateMesh.daeXUp (0 ms)
[----------] 5 tests from CreateMesh (2 ms total)
```

Tests pass, +1

I think this should be back-ported to I/J, also

de-vri-es:
Fixed code style issues in the test.

de-vri-es:
Also added a test to verify that `<unit>` tags are ignored.

gavanderhoorn:
Just noticed this PR. Nice catch @de-vri-es, but I'm a bit unsure about this part:

> One more important thing to note about this PR: it also undoes scaling from the `<unit>` metadata tag. Again, this is the same thing RViz does, and I think this behaviour matches expectations best, but it is an important detail nonetheless.

Units support in Collada is something that is frequently used, and without it, meshes that haven't been stored with their scaling set to meters will not load properly any more in ROS. Even if RViz does this, is it a good idea to deliberately remove support for a format feature like this?

Undoing/ignoring the root transform seems like a dubious thing to do as well -- at least to me. It'd be deliberately introducing behaviour that is different from the spec. Doesn't the fact that RViz does it make it essentially a buggy implementation of the Collada spec? @v4hn's comment in https://github.com/ros-planning/geometric_shapes/pull/52#issuecomment-242828595 would seem to apply here as well then?

A (seemingly) related issue: ros-visualization/rviz#853. Also: [Collada (.dae) meshes with unit != "meter" broken?](http://answers.ros.org/question/203188) and [Did RViz become unit-aware?](http://answers.ros.org/question/59514).

de-vri-es:
@gavanderhoorn: I understand your concern, and I'm also not entirely comfortable with clearing the root transform.

However, using the metadata tags only makes sense when you can specify matching parameters for the scene/world that the models are being imported in. If we have no knowledge of the scene/world we don't know how to incorporate the model metadata.

We could assume some set of standards for the scene/world: Z up and a 1 meter units. However, the 1 meter is not a official ROS standard. If I recall correctly, the advise is to use a unit such that "1" is as close as possible to size of your robot. (Actually, not any unit, but `meter` combined with an SI prefix.)

If no scene/world metadata is available, I do think that ignoring the metadata of the model is the best approach. Then at least a distance along an axis in the modelling software is the same distance along the same axis of the imported model.

Additionally, Assimp doesn't give us the tools required to convert to Z_UP reliably. It only transforms the model to Y_UP when a `<up_axis>` tag is present, and of course it leaves STL files alone. It also does not expose the loaded metadata.

We could extract the unit size from the root transform and only clear the rotation. But that does mean leaving a mismatch with RViz.

It may make more sense to first adjust `geometric_shapes` to match RViz and then raise this issue with RViz? On the other hand, that means multiple transitions for users of `geometric_shapes` if RViz decides to change.

de-vri-es:
On a side note: it would appear that Assimp can ignore the up direction for Collada files more cleanly by setting a property on the importer [1]:

```
importer->setPropertyInteger("IMPORT_COLLADA_IGNORE_UP_DIRECTION", 1)
```

That means we could still honor the unit tag easily (that is: even without manually decomposing the root transform and removing the rotation). I'm not convinced we should scale everything to meters though, unless there is a specification stating that `meters` are the units used throughout ROS?

[1] https://github.com/assimp/assimp/blob/master/code/ColladaLoader.cpp#L130

gavanderhoorn:
> However, using the metadata tags only makes sense when you can specify matching parameters for the scene/world that the models are being imported in. If we have no knowledge of the scene/world we don't know how to incorporate the model metadata.
> 
> We could assume some set of standards for the scene/world: Z up and a 1 meter units. However, the 1 meter is not a official ROS standard. If I recall correctly, the advise is to use a unit such that "1" is as close as possible to size of your robot.

To me, only 'meters' make sense as a scale for meshes. ROS uses meters throughout, and meters for meshes makes them align perfectly with TF and all of the other infrastructure that has a notion of distance embedded in it.

I can't find it right now, but I seem to remember that using meters for mesh scale is actually a (strong) recommendation.

> We could assume some set of standards for the scene/world: Z up and a 1 meter units

Afaik -- but I could've been living in my own private ROS universe :) -- that is already at least a convention. Units, chirality and coord orientation are defined in [REP-103](http://www.ros.org/reps/rep-0103.html). Perhaps not explicitly stated anywhere, but anything else in RViz, MoveIt or `geometric_shapes` makes little sense to me.

> It may make more sense to first adjust `geometric_shapes` to match RViz and then raise this issue with RViz? On the other hand, that means multiple transitions for users of `geometric_shapes` if RViz decides to change.

Hm. Well I don't think I can properly estimate the impact of such a change, but if something like this is changed, I'd be in favour of doing it once. And doing it right.

gavanderhoorn:
> I'm not convinced we should scale everything to meters though, unless there is a specification stating that meters are the units used throughout ROS?

Bit 'broken record', but [REP-103](http://www.ros.org/reps/rep-0103.html)?

de-vri-es:
Hmm, [REP-103](http://www.ros.org/reps/rep-0103.html) it is only informal and not very strongly worded, but that could still be enough justification to enforce Z up and 1 meter units. Although it would be nice if it can be disabled at the `geometric_shapes` level if people want to deviate from Z up or meter units.

However, the information Assimp gives is not enough to enforce Z up. Apparently gazebo decided to parse the collada files themselves to extract the `<up_axis>` tag. Not very pretty, but it doesn't look like Assimp gives us any choice if we want to enforce any convention.

That leaves 3 options:
1) Keep it as is, ignore both `<up_axis>` and `<unit>` metadata the way RViz currently does it.
2) Scale to meters using the `<unit>` metadata, ignore `<up_axis>`. A bit half-assed but doesn't require parsing the Collada file.
3) Scale to meters and enforce `Z_UP`. Requires parsing the Collada ourselves.

If we decide to go for 2 or 3 we should make sure to discuss with RViz so we end up with the same solution. In that case I would also like to make the use of the metadata tags optional or even configurable to other conventions in the `geometric_shapes` API (with default values that match Z up and meter units).

/edit: I think 2 is a pretty bad idea, btw

gavanderhoorn:
@de-vri-es wrote:

> Hmm, [REP-103](http://www.ros.org/reps/rep-0103.html) it is only informal and not very strongly worded, but that could still be enough justification to enforce Z up and 1 meter units

the only thing that is not explicit enough afaict is that the same conventions should be applied to / followed for mesh data / robot geometry. I would think the [Units](http://www.ros.org/reps/rep-0103.html#id16) and [Coordinate Frame Conventions](http://www.ros.org/reps/rep-0103.html#id19) sections pretty much explicitly cover the rest.

Regarding your suggestions: I think the important thing would be that we communicate whatever we end up choosing. Personally, I'd like units to be respected (if at all possible, but the Gazebo devs apparently found it too confusing / they found that modelling tools aren't consistent enough with it) and the model to be rotated so that Z is up (ie: rotate everything until the mesh-local coord frame is aligned with the ROS global coord frame.

Rotating objects like that isn't even very special: iirc, there are lots of tools that do that (game engines, viewers, etc).

But I really think this is something that could benefit from some more input. @wjwwood, @tfoote: comments?

v4hn:
If you like to see this fixed in _any_ way other than _exactly_ how RViz is doing it, please file an issue/pull-request against rviz first.

It is right, that we should respect the format specs. However, it looks like RViz (and gazebo?) decided to intentionally restrict the specs to a _subset_ of the file format.
As _everyone_ I know who ever used ROS at some point used RViz for visualization, they have all noticed potential differences and changed their files to adhere to the restricted specs.
So, very sadly, the specs for `geometric_shapes` to adhere to are not so much the specs for the file format, but more the restricted implementation by RViz.
(but yeah, you pretty much got me here @gavanderhoorn -.-)

de-vri-es:
> If you like to see this fixed in _any_ way other than _exactly_ how RViz is doing it, please file an issue/pull-request against rviz first.

I filed ros-visualization/rviz#1045. We should indeed match RViz, but it is worth waiting if RViz is going to change soon.

gavanderhoorn:
@gavanderhoorn wrote:

> Rotating objects like that isn't even very special: iirc, there are lots of tools that do that (game engines, viewers, etc).

Just as an example of this: the documentation on the Collada loader in the Torque 3D engine (from [here](http://docs.garagegames.com/torque-3d/official/content/documentation/Artist%20Guide/Formats/ColladaLoader.html)):

> **Up Axis**
> 
> The coordinate system in Torque 3D is equivalent to Z_UP, and the loader will automatically convert models using X_UP or Y_UP to this coordinate system. Some COLLADA exporters may generate models with a wrong or missing <up_axis> element (Z_UP is assumed if <up_axis> is not specified). In this case, the COLLADA import dialog can be used to override the value.

The same link also has a section on units.

davetcoleman:
I agree with @v4hn we should make geometric_shapes behave exactly as Rviz. Perhaps someone more familiar with Rviz could comment on the possibility of creating a shared code path for this functionality, e.g. Rviz depend on geometric_shapes or vice versa?

I doubt we will see any big changes like this to Rviz in the near future since it has such a HUGE user base. We will probably want to merge this PR

It would also be great if @de-vri-es or @gavanderhoorn could sum up this discussion into a new tutorial on how to use meshes (scaling, orientation, file format) with MoveIt! in [moveit_tutorials](http://docs.ros.org/indigo/api/moveit_tutorials/html/index.html)

v4hn:
@de-vri-es I'm pretty certain this will not change in rviz kinetic anymore.
Is this request in your opinion ready to merge into i/j/k?

de-vri-es:
You're probably right. No reply yet at ros-visualization/rviz#1045.

It is ready to merge I think. The unit tests are in place to verify the behaviour both for `<up_axis>` and `<unit>`.

v4hn:
ok. +1 as is. This should be merged to indigo and jade too.

davetcoleman:
thanks @de-vri-es. cherry-picked to indigo

PR #59, opened 2016-09-30
Fix for linking eigen3

nikonikolov:
Added the following fix so that pkgconfig is used to find eigen3. Not having this caused an error for me while I was building ros from source on Slackware Linux 14.2:

```
CMake Error at CMakeLists.txt:24 (find_package):
  By not providing "FindEigen3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Eigen3", but
  CMake did not find one.

  Could not find a package configuration file provided by "Eigen3" with any
  of the following names:

    Eigen3Config.cmake
    eigen3-config.cmake

  Add the installation prefix of "Eigen3" to CMAKE_PREFIX_PATH or set
  "Eigen3_DIR" to a directory containing one of the above files.  If "Eigen3"
  provides a separate development package or SDK, be sure it has been
  installed.
```

Hope this does not mess up with other distros.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,9 @@ find_package(Boost REQUIRED system filesystem)
 
 find_package(console_bridge REQUIRED)
 
-find_package(Eigen3 REQUIRED)
+find_package( PkgConfig )
+pkg_check_modules( EIGEN3 REQUIRED eigen3 )
+include_directories( ${EIGEN3_INCLUDE_DIRS} ) 
 
 find_package(octomap REQUIRED)
 


v4hn:
I'm inclined to close this as "won't fix".
- Many other packages in ROS require cmake support for Eigen3, so this is only one of quite a number of places that will fail
- This is (arguably) a bug, or at least a short-coming, in Slackware's eigen3 package: Eigen3 ships a `cmake/FindEigen3.cmake` in its release tarball for quite a number of releases now. Most (and I mean "most" and not just the major ones) distros install this file into cmake's `Modules` folder, thus providing Eigen3 cmake-support. Slackware does not ship that file
- Eigen3.3, which is about to be released (although debian uses the beta release for quite some time already), provides proper cmake config files, so it is quite a big step back to revert to the pkgconfig.

nikonikolov:
Fair enough, we can agree this is slackware specific then.

v4hn:
ok, closing.

PR #72, opened 2017-10-20
Fix log function deprecation warnings.

### LINKED ISSUE #71: Use of deprecated logging functions
See: https://github.com/ros/console_bridge/blob/master/include/console_bridge/console.h#L68-L90

https://github.com/ros-planning/geometric_shapes/search?q=logWarn

https://github.com/ros-planning/geometric_shapes/search?q=logError

mikepurvis:
Closes #71.

### DIFF
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -788,7 +788,7 @@ void bodies::ConvexMesh::useDimensions(const shapes::Shape *shape)
 
   if (exitcode != 0)
   {
-    logWarn("Convex hull creation failed");
+    CONSOLE_BRIDGE_logWarn("Convex hull creation failed");
     qh_freeqhull (!qh_ALL);
     int curlong, totlong;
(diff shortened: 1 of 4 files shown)


racko:
Since the release of console_bridge 0.4.0 (https://github.com/ros/console_bridge/commit/c227ae01ce16d73facb477644d9c25a5ddb166fa), these have turned from deprecation warnings into build errors:

```
.../geometric_shapes-release-release-lunar-geometric_shapes-0.5.3-0/src/body_operations.cpp:62:7: error: ‘logError’ was not declared in this scope
       logError("Creating body from shape: Unknown shape type %d", (int)shape->type);
```

Also, it seems that the change needs an update since your change does not touch body_operations.cpp, which has a number of `logError` occurrences.

davetcoleman (approved):
Thanks for fixing this regression

PR #79, opened 2018-04-06
add operator<< for enum

v4hn:
Just some micro-additions && simplifications

### DIFF
--- a/src/shapes.cpp
+++ b/src/shapes.cpp
@@ -47,6 +47,33 @@ const std::string shapes::Mesh::STRING_NAME = "mesh";
 const std::string shapes::Plane::STRING_NAME = "plane";
 const std::string shapes::OcTree::STRING_NAME = "octree";
 
+namespace shapes {
+std::ostream& operator<<(std::ostream& ss, ShapeType type)
+{
+   switch(type){
+      case UNKNOWN_SHAPE:
+         ss << "unknown"; break;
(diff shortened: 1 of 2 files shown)


src/shapes.cpp
```diff
@@ -39,76 +39,103 @@
...
-const std::string shapes::Cylinder::STRING_NAME = "cylinder";
-const std::string shapes::Cone::STRING_NAME = "cone";
-const std::string shapes::Mesh::STRING_NAME = "mesh";
-const std::string shapes::Plane::STRING_NAME = "plane";
-const std::string shapes::OcTree::STRING_NAME = "octree";
-
-shapes::Shape::Shape()
+namespace shapes {
+
+const std::string Sphere::STRING_NAME = "sphere";
```
bmagyar, line 44: How about using a `static const std::map<ShapeType, std::string>` ?

C++11 allows for this:
```
std::map <int, std::string> x { 
  { 42, "foo" }, 
  { 3, "bar" } 
};
```

v4hn: Sure, we can do that. But the static `STRING_NAME` variables are - ridiculous enough - public API, although not even virtual. So if you like to see this changed please create a melodic branch and file a request there.
I wanted to retain API here to merge into kinetic.

rhaschke (approved):

PR #95, opened 2018-12-10
Simplify Travis tests

rhaschke:
- only check clang-format once
- only build once against full MoveIt workspace


v4hn:
makes sense.

PR #97, opened 2019-02-21
Made Body::containsPoint() consistent in surface points inclusion. (melodic)

peci1:
Retargeted #90 to melodic-devel.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,0 +1,7 @@
+# Migration Notes
+
+API changes in geometric_shapes releases
+
+## ROS Melodic
+
+- There was a change in point containment computations - now all bodies are expected to also contain their surface points. [The old behavior was inconsistent](https://github.com/ros-planning/geometric_shapes/issues/91). Rounding errors may still yield surprising results, though.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -13,6 +13,8 @@ Supported shapes:
 
 Note: Bodies for meshes compute the convex hull of those meshes in order to provide the point containment / ray intersection routines.
 
+Note: There was a recent change in point containment computations - now all bodies are expected to also contain their surface points. [The old behavior was inconsistent](https://github.com/ros-planning/geometric_shapes/issues/91). Rounding errors may still yield surprising results, though.
```
rhaschke: This should be moved into MIGRATION.md.

rhaschke (changes requested):
Thanks for retargeting this to Melodic. Could please squash the last two commits (fix formatting).

peci1:
Squashed and created MIGRATION.md (there was none before).

rhaschke:
> Squashed and created MIGRATION.md (there was none before).

Ah, sorry, we have a MIGRATION.md only for the main MoveIt repository.
I'm not sure, it's worth to mention these changes in README.md or a new MIGRATION.md.
Usually such (minor) changes are reported in the ChangeLog only. What do you think?

peci1:
Well, the newly created MIGRATION.md could also contain information about the merge with shape_tools... I think it might come handy...

PR #99, opened 2019-03-20
[Windows][melodic] Add ASSIMP_LIBRARY_DIRS for Windows build.

seanyen:
Add ASSIMP_LIBRARY_DIRS for Windows build.

(Verified in https://aka.ms/ros project.)

PR #101, opened 2019-04-01
[Windows][melodic] Fix ASSIMP libraries path for Windows build

seanyen:
This change is another iteration based on #99.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,8 @@ if (NOT ASSIMP_FOUND)
 endif()
 
 # find *absolute* paths to ASSIMP_LIBRARIES
+# Both, pkg-config and assimp's cmake-config don't provide an absolute library path.
+# For, pkg-config the path is in ASSIMP_PC_LIBRARY_DIRS, for cmake in ASSIMP_LIBRARY_DIRS.
 find_library(ASSIMP_ABS_LIBRARIES NAMES ${ASSIMP_LIBRARIES} assimp HINTS ${ASSIMP_LIBRARY_DIRS} ${ASSIMP_PC_LIBRARY_DIRS})
 set(ASSIMP_LIBRARIES "${ASSIMP_ABS_LIBRARIES}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,10 +15,12 @@ if (NOT ASSIMP_FOUND)
   find_package(PkgConfig REQUIRED)
   # assimp is required, so REQUIRE the second attempt
   pkg_check_modules(ASSIMP_PC REQUIRED assimp)
-  # find *absolute* paths to ASSIMP_LIBRARIES
   set(ASSIMP_INCLUDE_DIRS ${ASSIMP_PC_INCLUDE_DIRS})
-  find_library(ASSIMP_LIBRARIES assimp HINTS ${ASSIMP_LIBRARY_DIRS})
 endif()
+
+# find *absolute* paths to ASSIMP_LIBRARIES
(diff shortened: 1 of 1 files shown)


seanyen:
@rhaschke Sorry I wasn't able to find time to iterate on #99 and here is a version based on your change to make it work on Windows build.

CMakeLists.txt
```diff
@@ -18,6 +18,9 @@ if (NOT ASSIMP_FOUND)
   # find *absolute* paths to ASSIMP_LIBRARIES
   set(ASSIMP_INCLUDE_DIRS ${ASSIMP_PC_INCLUDE_DIRS})
   find_library(ASSIMP_LIBRARIES assimp HINTS ${ASSIMP_LIBRARY_DIRS})
```
rhaschke: Please try this instead:
```suggestion
   find_library(ASSIMP_LIBRARIES assimp HINTS ${ASSIMP_PC_LIBRARY_DIRS})
```

rhaschke (changes requested):
Looks like I missed to add `PC`. Please try this instead of your addition.

seanyen:
@rhaschke Hi, it turned out on Windows build, `assimp` is discovered by using `CMake` config file so the `pkg-config` code path is not really getting exercised. Anyway, I am pushing a new change which should accommodate both. Let me know what you think.

rhaschke:
Ok, if Windows detects assimp via a cmake config, what is the actual problem. Shouldn't this work out of the box?

seanyen:
On Windows build, we are using `assimp` v4.1 (the same like Ubuntu bionic using) and just as many libraries they are producing the CMake config intead of `pkg-config` for Windows build. And from `assimp`, its CMake Config is like this: https://github.com/assimp/assimp/blob/v4.1.0/assimp-config.cmake.in#L42-L51

As you can see, the `${ASSIMP_LIBRARIES}` is defined to be just as the library name like `assimp-vc140-mt` and the library directory is defined in another CMake variable `${ASSIMP_LIBRARY_DIRS}`. However, `${ASSIMP_LIBRARY_DIRS}` was not in the target's link directory, so duing linking stage, it cannot find where the library actually is.

CMakeLists.txt
```diff
@@ -15,11 +15,13 @@ if (NOT ASSIMP_FOUND)
   find_package(PkgConfig REQUIRED)
   # assimp is required, so REQUIRE the second attempt
   pkg_check_modules(ASSIMP_PC REQUIRED assimp)
-  # find *absolute* paths to ASSIMP_LIBRARIES
   set(ASSIMP_INCLUDE_DIRS ${ASSIMP_PC_INCLUDE_DIRS})
-  find_library(ASSIMP_LIBRARIES assimp HINTS ${ASSIMP_LIBRARY_DIRS})
 endif()
 
+# find *absolute* paths to ASSIMP_LIBRARIES
```
rhaschke, line 21:
```suggestion
# find *absolute* paths to ASSIMP_LIBRARIES
# Both, pkg-config and assimp's cmake-config don't provide an absolute library path.
# For, pkg-config the path is in ASSIMP_PC_LIBRARY_DIRS, for cmake in ASSIMP_LIBRARY_DIRS.
```

rhaschke (changes requested):
I added a comment. Otherwise lgtm.

seanyen:
@rhaschke Thanks. The comments are added.

rhaschke (approved):

henningkayser (approved):
Looks good to me!
The kinetic build timed out on Travis so let's wait for the new run to finish.

PR #104, opened 2019-05-22
Added bodies::Body::computeBoundingBox (aligned box version).

peci1:
Solves #89 for axis-aligned bounding boxes.

PR #105, opened 2019-05-24
Added bodies::Body::computeBoundingBox (oriented box version). 

peci1:
Added support for oriented bounding boxes. Depends on #104.

Please note that this PR adds dependency on FCL. I checked how many dependees of geometric_shapes there are which do not depend on fcl, and only collada_urdf seems to get an unnecessary transitive dependency on FCL. moveit_core depends on FCL.

The OBBs for single bodies are all the tightest you can get (they have some degrees of freedom though, e.g. for sphere and cylinder). For spheres, the axis-aligned BB is returned. For cylinder, the BB rotates around z axis with the cylinder. The mesh OBB can be suboptimal since it is initialized with the AABB.

Once you start merging OBBs, the result is only an approximate OBB (which is what FCL computes). As I get it, the optimal (minimum volume) OBB computation is heavy and is tackled using iterative methods in other libraries. I haven't yet implemented the optimal version, and I probably even won't do it. However, if you think it might be valuable in the future, we can add the function header and leave it unimplemented (or throwing an exception/assert).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,8 +49,13 @@ find_package(Eigen3 REQUIRED)
 find_package(octomap REQUIRED)
 
 pkg_check_modules(LIBFCL_PC REQUIRED fcl)
+# find *absolute* paths to LIBFCL_* paths
 find_path(LIBFCL_INCLUDE_DIRS fcl/config.h HINTS ${LIBFCL_PC_INCLUDE_DIR} ${LIBFCL_PC_INCLUDE_DIRS})
-find_library(LIBFCL_LIBRARIES fcl HINTS ${LIBFCL_PC_LIBRARY_DIRS})
+set(LIBFCL_LIBRARIES)
+foreach(_lib ${LIBFCL_PC_LIBRARIES})
+  find_library(_lib_${_lib} ${_lib} HINTS ${LIBFCL_PC_LIBRARY_DIRS})
+  list(APPEND LIBFCL_LIBRARIES ${_lib_${_lib}})
+endforeach()
 
 find_package(catkin REQUIRED COMPONENTS
   eigen_stl_containers
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,9 +21,14 @@ if (NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
+# used for finding Assimp and FCL
+find_package(PkgConfig REQUIRED)
+
+# TODO(henningkayser): Remove policy fix when assimp 5.1 is available
+# Suppress policy warning in assimp (https://github.com/assimp/assimp/pull/2722)
+set(CMAKE_POLICY_DEFAULT_CMP0012 NEW)
 find_package(ASSIMP QUIET)
-if (NOT ASSIMP_FOUND)
-  find_package(PkgConfig REQUIRED)
+if(NOT ASSIMP_FOUND)
   # assimp is required, so REQUIRE the second attempt
   pkg_check_modules(ASSIMP_PC REQUIRED assimp)
   set(ASSIMP_INCLUDE_DIRS ${ASSIMP_PC_INCLUDE_DIRS})
@@ -42,6 +47,10 @@ find_package(console_bridge REQUIRED)
 find_package(Eigen3 REQUIRED)
 
 find_package(octomap REQUIRED)
+
+pkg_check_modules(LIBFCL_PC REQUIRED fcl)
+find_path(LIBFCL_INCLUDE_DIRS fcl/config.h HINTS ${LIBFCL_PC_INCLUDE_DIR} ${LIBFCL_PC_INCLUDE_DIRS})
+find_library(LIBFCL_LIBRARIES fcl HINTS ${LIBFCL_PC_LIBRARY_DIRS})
 
 find_package(catkin REQUIRED COMPONENTS
   eigen_stl_containers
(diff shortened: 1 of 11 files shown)


rhaschke:
I'm hesitant to include the FCL dependency in this basic lib.
Could we integrate such code as a utility lib in MoveIt instead?

peci1:
Another option would be to completely get rid of FCL here and only provide OBBs for basic shapes. Merging OBBs would be left to the user. As there are multiple ways to do it, it seems even like the best solution to me.

peci1:
On the other hand, I actually wonder this library doesn't depend on FCL. Maybe that could lead to some code cleanup. But the only version that makes sense (since it uses Eigen types) would be 0.6, which has been released on Feb 2020 and I thus don't think this version will be available in Focal... Older versions of Ubuntu are out of question (or a custom build of FCL would need to be run, which I can imagine could be a big burden).

CMakeLists.txt
```diff
@@ -61,20 +66,23 @@ if (HAVE_QHULL_2011)
 endif()
 
 include_directories(include)
+include_directories(SYSTEM ${LIBFCL_INCLUDE_DIRS})
```
mamoll: Why `SYSTEM`?

src/aabb.cpp
```diff
@@ -0,0 +1,64 @@
...
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+/* Author: Martin Pecka */
+
+#include <geometric_shapes/aabb.h>
+#include <fcl/shape/geometric_shapes_utility.h>
```
mamoll: remove/comment out since FCL isn't actually used?

DLu:
@peci1 Can you restage this to resolve the conflicts, now that #104 is resolved?

src/body_operations.cpp
```diff
@@ -175,3 +175,15 @@ void bodies::computeBoundingSphere(const std::vector<const bodies::Body*>& bodie
...
+{
+  for (const auto& box : boxes)
+    mergedBox.extend(box);
+}
+
+void bodies::mergeBoundingBoxesApprox(const std::vector<bodies::OBB>& boxes, bodies::OBB& mergedBox)
+{
+  for (const auto& box : boxes)
+    mergedBox.extendApprox(box);
+}
```
mamoll: add new line

src/obb.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+bool OBB::overlaps(const bodies::OBB& other)
+{
+  return obb_->overlap(*other.obb_);
+}
+
+OBB::~OBB() = default;
+
+}
```
mamoll: add new line

mamoll (changes requested):
Looks good overall. Minor changes needed. Perhaps an issue should be created to track when FCL min. version is bumped to 0.6, so that we don't forget to simplify code (perhaps moveit's fcl code could also be simplified?

peci1:
Hi, I'm sorry, I forgot a bit about this PR. I rebased and cleaned up the formatting.

src/obb.cpp
peci1: fixed

src/body_operations.cpp
peci1: fixed

src/aabb.cpp
peci1: moved to a TODO comment

CMakeLists.txt
peci1: So that errors and warnings internal to FCL do not spam build output of geometric_shapes.

peci1:
I'm not sure if it's okay to add the FCL dependency. If the package.xml switched to format 2, it could be made only a build and exec dependency and not a build-export dependency. Which means downstream packages should not be affected in their build, but FCL would anyways need to be installed. According to my "qualified guess", the number of installations of geometric_shapes which do not have FCL would be almost zero. But if it is a problem, I can make it an optional dependency (disabling the oriented bounding box functionality if FCL is not found). So binary installations would get FCL as a part of the normal update process, while local source installations would continue to work even if FCL would not get installed.

simonschmeisser:
there is a `noetic-devel` branch by now and `ros-noetic-fcl` is at 0.6.1 so it might make sense to retarget and update the branch?

peci1:
`noetic-devel` continuation in #210 . I also figured out this PR breaks ABI by adding a virtual function. I worked around that in #210. So let's first solve the noetic PR and then see whether this PR could be used as an easy already prepared backport.

peci1:
With the noetic version already merged (#210), I helped a bit backporting this to melodic by cherry-picking the noetic commits. They apply cleanly, with just a small merge conflict in CMakeLists.txt I've resolved.

Could @moveit-maintainers please have a look?

simonschmeisser:
In #210 the ABI compatibility hack/fix commit was dropped, I'm not sure it's fair to do the same in melodic? @rhaschke might have a stronger opinion on it

peci1:
Melodic ABI has anyways be touched, so I think it should be okay to touch it here, too? The cherry-picked commits already do not contain do ABI compatibility commit.

rhaschke:
Regarding ABI compatibility, @v4hn recently argued we can drop it, because we use SONAME versioning of libs anyway.

rhaschke (approved):
LGTM. I pushed a minor change though: Formally the FCL version check should check for `<= 0.5`

peci1:
@simonschmeisser could you please have a look at this again and consider giving the second approval?

peci1:
With Melodic EOL around the corner, can I please ask the maintainers to look at this PR again? Thanks.

peci1:
@simonschmeisser [We're just 3 syncs away from Melodic EOL](https://discourse.ros.org/chat/c/-/11/490). Could you please (or any other maintainer) have a look at this PR?

AndyZe:
It looks like you addressed all of @mamoll's requested changes so I'll clear his review.

include/geometric_shapes/obb.h
```diff
@@ -0,0 +1,132 @@
...
+   * \param [in] pose New pose of the OBB.
+   * \param [in] extents New extents of the OBB.
+   */
+  void setPoseAndExtents(const Eigen::Isometry3d& pose, const Eigen::Vector3d& extents);
+
+  /**
+   * \brief Get the extents of the OBB.
+   * \return The extents.
+   */
+  Eigen::Vector3d getExtents() const;
```
AndyZe, line 73: If I wanted to be extremely nitpicky, `[[nodiscard]]` could be added here. Although I think it's fine, especially since this was already merged to Noetic.

AndyZe (approved):
Looks fine to me, especially considering the Noetic version was already merged. I'll create an issue to update this code when FCL v0.6 is available, as Mark suggested.

peci1:
Thanks for merging!

include/geometric_shapes/obb.h
peci1: `[[nodiscard]]` is not very well handled by gcc (maybe just the older versions, I don't know exactly). I remember I once tried to add it to some methods and then removed it again because of these issues...

peci1:
@AndyZe any idea when a new Melodic release is planned? There are 3 Melodic syncs left, and if I want to base some changes in other repos depending on this PR, the timing is getting tight.

AndyZe:
I don't have any special information. I guess you saw this post:

https://discourse.ros.org/t/new-packages-for-melodic-2023-03-02/30071

peci1:
Yes, I'm aware of that post, especially:

>  If there are updates you want to get into it, now would be a good time to release them.

Maybe there was a misunderstanding. I was asking for when a next version of geometric_shapes will be released.

AndyZe:
I see. Maybe @vatanaksoytezer @tylerjw or @henningkayser could make a release soon

PR #106, opened 2019-05-28
Throw runtime exception when a shape or body should have a negative dimension.

peci1:
Meshes are exceptions - you might want to give it a negative scale to mirror it. Although I'm not sure all member functions would tackle negative scale or large negative padding correctly.

### DIFF
--- a/src/shapes.cpp
+++ b/src/shapes.cpp
@@ -101,6 +101,8 @@ Sphere::Sphere() : Shape()
 
 Sphere::Sphere(double r) : Shape()
 {
+  if (r < 0)
+    throw std::runtime_error("Sphere radius must be non-negative.");
   type = SPHERE;
   radius = r;
 }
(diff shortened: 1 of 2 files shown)


peci1:
I was using random() to generate shapes and negative dimensions slipped my attention. What's interesting is they got through the whole pipeline of shape, body, marker and they even displayed in rviz. But the contains/intersectsRay tests are wrong for such bodies.

src/shapes.cpp
```diff
@@ -285,18 +293,24 @@ void Shape::padd(double padding)
 void Sphere::scaleAndPadd(double scale, double padding)
 {
   radius = radius * scale + padding;
+  if (radius < 0)
```
rhaschke: I think the check for valid dimensions should be made before modifying the object.

rhaschke (changes requested):
I agree that we should check for negative dimensions. In the constructors the checks are fine already.
In the other methods, you should perform the check _before_ modifying the object. Otherwise you leave the object in an undefined state - even if the exception is correctly catched and handled by the caller.

peci1:
I did the requested fixes, please, check if the PR is ready now.

rhaschke (commented):
Code looks good. Please rebase onto melodic-devel. We can't accept merge commits in PRs.

rhaschke (approved):

peci1:
Okay, rebased.

v4hn:
Merging with second approval.
Thanks @peci1 .

PR #107, opened 2019-05-28
Corrected bounding cylinder computations

peci1:
The computation of bounding cylinder of a cylinder was wrong - padding has to be added twice to the length of the cylinder. Exchanged for `2 * length2_`, where `length2_` is a an internal data member which contains half of the scaled length plus once padding. Thus multiplying this member by two, we get exactly `scale_ * length_ + 2 * padding_`.

Also, computation of bounding cylinder of a sphere was wrong, because length of the cylinder has to be twice the radius of it.

Computation of bounding cylinder of a ConvexMesh was also wrong, it wasn't taking into account scaling or padding completely.

### DIFF
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -180,7 +180,7 @@ void bodies::Sphere::computeBoundingCylinder(BoundingCylinder& cylinder) const
 {
   cylinder.pose = pose_;
   cylinder.radius = radiusU_;
-  cylinder.length = radiusU_;
+  cylinder.length = 2.0 * radiusU_;
 }
 
 void bodies::Sphere::computeBoundingBox(bodies::AABB& bbox) const
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -1071,8 +1071,9 @@ void bodies::ConvexMesh::computeBoundingSphere(BoundingSphere& sphere) const
 
 void bodies::ConvexMesh::computeBoundingCylinder(BoundingCylinder& cylinder) const
 {
-  cylinder.length = mesh_data_ ? mesh_data_->bounding_cylinder_.length : 0.0;
-  cylinder.radius = mesh_data_ ? mesh_data_->bounding_cylinder_.radius : 0.0;
+  // the padding contibution might be smaller in reality, but we want to get it right for the worst case
+  cylinder.length = mesh_data_ ? mesh_data_->bounding_cylinder_.length * scale_ + 2 * padding_ : 0.0;
+  cylinder.radius = mesh_data_ ? mesh_data_->bounding_cylinder_.radius * scale_ + padding_ : 0.0;
   // need to do rotation correctly to get pose, which bounding box does
   BoundingCylinder cyl;
   bounding_box_.computeBoundingCylinder(cyl);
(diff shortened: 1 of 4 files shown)


peci1:
Beware: this PR depends on #106. If you don't want it, cherry-pick instead.

rhaschke (changes requested):
The commit 8c402215f50122f3bc4b9effaa4fd17ec28755d4 looks good, but the PR also comprises #106. This should be factored out.
Please provide a description as well.

peci1:
Ok, I made this PR independent from the others and added description.

rhaschke:
@peci1, thanks for adding a description. However, the description doesn't fit to your change: You actually removed the padding from the computation.

Thanks for resolving the conflict. However, may I ask you to _rebase_ this PR onto `melodic-devel`. We don't accept PRs that contain merge commits.

rhaschke (approved):

peci1:
I reworked this PR, also adding a test suite. The tests revealed more bugs in sphere and convex mesh, too, so I fixed them.

peci1:
And I suspect not all functions in ConvexMesh work correctly with scale/padding. But I don't have the resources to write the test cases and fix it...

src/bodies.cpp
```diff
@@ -1071,8 +1071,8 @@ void bodies::ConvexMesh::computeBoundingSphere(BoundingSphere& sphere) const
 
 void bodies::ConvexMesh::computeBoundingCylinder(BoundingCylinder& cylinder) const
 {
-  cylinder.length = mesh_data_ ? mesh_data_->bounding_cylinder_.length : 0.0;
-  cylinder.radius = mesh_data_ ? mesh_data_->bounding_cylinder_.radius : 0.0;
+  cylinder.length = mesh_data_ ? mesh_data_->bounding_cylinder_.length * scale_ + 2 * padding_ / sqrt(3) : 0.0;
+  cylinder.radius = mesh_data_ ? mesh_data_->bounding_cylinder_.radius * scale_ + padding_ / sqrt(3) : 0.0;
```
v4hn: I did not look through the whole patch here, so just one comment here:

Assuming this patch is correct, I believe it is justified to add a comment explaining where the `sqrt(3)` comes from.

peci1: Learning more about the mesh scaling in my other PRs, I know see that the sqrt(3) shouldn't be here. The idea was that if the padding is applied along a diagonal, its contribution to each axis is padding/sqrt(3). However, the "worst case" here is that the "maximal" vertex is scaled along an axis, and the it will get the full padding. It can thus happen that the bounding cylinder will not be as tight as it can be, but well, for the meshes, there are more problems around, and even the bounding box computed internally by ConvexMesh is not tight (it uses the same logic as I described in this comment).

peci1: For meshes without padding, the bounding cylinder will be tight.

peci1:
Okay, I've updated the code, actually deleting the cryptic `sqrt(3)` as it shouldn't really be there. I've updated tests to reflect the fact that bounding cylinders of padded meshes are not tight.

There might be some method that would compute tight cylinders in all cases, but I'd leave that on another PR after it's decided how padding in meshes should be treated.

v4hn (approved):
Thank you for digging through this mess fixing the computations. :sunflower:

v4hn:
explicitly merged via merge commit.

PR #108, opened 2019-05-28
Fixed Box::intersectsRay to return both intersections.

peci1:
Box::intersectsRay was always returning only the first collision when  `collision` arg was set to zero.
    
This is wrong according to function docs, so I corrected that.

### DIFF
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -700,7 +700,7 @@ bool bodies::Box::intersectsRay(const Eigen::Vector3d& origin, const Eigen::Vect
     if (tmax - tmin > detail::ZERO)
     {
       intersections->push_back(tmin * dir + origin);
-      if (count > 1)
+      if (count == 0 || count > 1)
         intersections->push_back(tmax * dir + origin);
     }
     else


peci1:
Beware: this PR depends on #106 and #107. If you don't want them all, cherry-pick instead.

src/bodies.cpp
```diff
@@ -652,7 +661,7 @@ bool bodies::Box::intersectsRay(const Eigen::Vector3d& origin, const Eigen::Vect
     if (tmax - tmin > detail::ZERO)
     {
       intersections->push_back(tmin * dir + origin);
-      if (count > 1)
+      if (count == 0 || count > 1)
```
rhaschke, line 703:
```suggestion
      if (count >= 1)
```

rhaschke (changes requested):
Approving the change semantically. Please consider the suggested simplification.
Please factor out the last commit for this PR. The other commits are handled in other PRs.

src/bodies.cpp
peci1: I don't think your suggestion is correct... maybe `count != 1` would look better, but I thought writing it down explicitly like this would be less confusing for future readers of the code

peci1:
I made this PR independent from the the others.

src/bodies.cpp
rhaschke: You are right, I misread the condition. Maybe, indeed `count != 1` would be the clearer expression.

rhaschke: What's the meaning of `count` here? I thought it would be "number of intersections".

rhaschke:
> I made this PR independent from the the others.

I don't understand, why you merged all your pending PRs with the melodic-devel branch.
It's strongly discouraged to have merge commits inside a PR branch, because this will mess up the commit history when merging via github merge. Please fix this.

rhaschke (approved):

src/bodies.cpp
peci1: it's not an output parameter, but an input one - if set to 1, it tells the algorithm it doesn't need to spend time searching for the second intersection if there is some... 

I still think that the way I put the condition is better self-documenting as it explicitly says that in case `count == 0` we also want to process all the intersections

peci1:
Could somebody please give this PR the second approval? I think it should be ready to be merged.

peci1:
I fixed the history and rebased onto current melodic-devel.

The merges with melodic-devel appeared as a part of my workflow - checkout branch, merge it with upstream, continue changes. Rebasing is another alternative, and probably somehow clearer from the point of view of the project history. But it requires force-pushing, which I generally don't like...

v4hn (approved):
makes sense. merging.

PR #109, opened 2019-05-28
Fixed computation of intersectsRay() for Cylinder, Box and ConvexMesh

peci1:
This PR fixes a multitude of problems the current implementations of `intersectsRay()` had.

In Box, even according to the link to the used method, the approach should only work for axis-aligned boxes. But bodies::Box can be arbitrarily transformed. So I instead strip rotation from both the box and the ray, which gives the same results. The old implementation would fail every time the untransformed min corner ceased to be the min corner of the transformed box.

For Cylinder, there were a few cornercases when incorrect number of intersections were returned (e.g. when the ray is tangential to the cylinder or when it goes through the edge between base and sides).

For ConvexMesh, there were several problems. Many of them arose from a wrong approach to incorporate padding. Also, if a ray hit the edge of two triangles, it was reported twice.

I also restructured the tests for both `containsPoint()` and `intersectsRay()` and added a lot of new tests - both randomized and testing corner cases. There is also a regression test `BoxRayIntersection::Regression109` which fails on current melodic-devel and succeeds with this test.

Aside from the fixes in `intersectsRay()`, I also fixed `Cylinder::samplePointInside()` and `ConvexMesh::containsPoint()` whose correct working is needed by the unit tests.

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -496,12 +496,10 @@ protected:
 
   // pose/padding/scaling-dependent values & values computed for convenience and fast upcoming computations
   Eigen::Vector3d center_;
-  Eigen::Vector3d normalL_;
-  Eigen::Vector3d normalW_;
-  Eigen::Vector3d normalH_;
+  Eigen::Matrix3d invRot_;
 
-  Eigen::Vector3d corner1_;
-  Eigen::Vector3d corner2_;
+  Eigen::Vector3d corner1_;  //!< The translated, but not rotated min corner
+  Eigen::Vector3d corner2_;  //!< The translated, but not rotated max corner
 
   double length2_;
   double width2_;
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -327,7 +327,7 @@ public:
   {
     type_ = shapes::SPHERE;
     shapes::Sphere shape(sphere.radius);
-    setDimensions(&shape);
+    setDimensionsDirty(&shape);
 
     Eigen::Isometry3d pose = Eigen::Isometry3d::Identity();
     pose.translation() = sphere.center;
@@ -389,7 +389,7 @@ public:
   {
     type_ = shapes::CYLINDER;
     shapes::Cylinder shape(cylinder.radius, cylinder.length);
-    setDimensions(&shape);
+    setDimensionsDirty(&shape);
     setPose(cylinder.pose);
   }
 
@@ -458,7 +458,7 @@ public:
   {
     type_ = shapes::BOX;
     shapes::Box shape(aabb.sizes()[0], aabb.sizes()[1], aabb.sizes()[2]);
-    setDimensions(&shape);
+    setDimensionsDirty(&shape);
 
     Eigen::Isometry3d pose = Eigen::Isometry3d::Identity();
     pose.translation() = aabb.center();
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -569,7 +569,12 @@ protected:
   /** \brief (Used mainly for debugging) Count the number of vertices behind a plane*/
   unsigned int countVerticesBehindPlane(const Eigen::Vector4f& planeNormal) const;
 
-  /** \brief Check if a point is inside a set of planes that make up a convex mesh*/
+  /** \brief Check if the point is inside all halfspaces this mesh consists of (mesh_data_->planes_).
+   *
+   * \note The point is expected to have pose_ "cancelled" (have inverse pose of this mesh applied to it).
+   * \note Scale and padding of the mesh are taken into account.
+   * \note There is a 1e-9 margin "outside" the planes where points are still considered to be inside.
+   */
   bool isPointInsidePlanes(const Eigen::Vector3d& point) const;
 
   struct MeshData
@@ -578,6 +583,7 @@ protected:
     EigenSTL::vector_Vector3d vertices_;
     std::vector<unsigned int> triangles_;
     std::map<unsigned int, unsigned int> plane_for_triangle_;
+    std::map<unsigned int, unsigned int> triangle_for_plane_;
     Eigen::Vector3d mesh_center_;
     double mesh_radiusB_;
     Eigen::Vector3d box_offset_;
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -496,7 +496,9 @@ protected:
 
   // pose/padding/scaling-dependent values & values computed for convenience and fast upcoming computations
   Eigen::Vector3d center_;
-  Eigen::Matrix3d invRot_;
+  Eigen::Vector3d normalL_;
+  Eigen::Vector3d normalW_;
+  Eigen::Vector3d normalH_;
 
   Eigen::Vector3d corner1_;  //!< The translated, but not rotated min corner
   Eigen::Vector3d corner2_;  //!< The translated, but not rotated max corner
@@ -528,9 +530,7 @@ public:
     setDimensions(shape);
   }
 
-  virtual ~ConvexMesh()
-  {
-  }
+  virtual ~ConvexMesh();
 
   /** \brief Returns an empty vector */
   virtual std::vector<double> getDimensions() const;
@@ -583,7 +583,6 @@ protected:
     EigenSTL::vector_Vector3d vertices_;
     std::vector<unsigned int> triangles_;
     std::map<unsigned int, unsigned int> plane_for_triangle_;
-    std::map<unsigned int, unsigned int> triangle_for_plane_;
     Eigen::Vector3d mesh_center_;
     double mesh_radiusB_;
     Eigen::Vector3d box_offset_;
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -530,17 +530,8 @@ bool bodies::Box::samplePointInside(random_numbers::RandomNumberGenerator& rng, 
 
 bool bodies::Box::containsPoint(const Eigen::Vector3d& p, bool /* verbose */) const
 {
-  Eigen::Vector3d v = p - center_;
-  double pL = v.dot(normalL_);
-  if (fabs(pL) > length2_)
-    return false;
-
-  double pW = v.dot(normalW_);
-  if (fabs(pW) > width2_)
-    return false;
-
-  double pH = v.dot(normalH_);
-  return fabs(pH) <= height2_;
+  const Eigen::Vector3d aligned = (invRot_ * (p - center_)).cwiseAbs();
+  return aligned[0] <= length2_ && aligned[1] <= width2_ && aligned[2] <= height2_;
 }
 
 void bodies::Box::useDimensions(const shapes::Shape* shape)  // (x, y, z) = (length, width, height)
@@ -580,12 +571,10 @@ void bodies::Box::updateInternalData()
   radiusB_ = sqrt(radius2_);
 
   ASSERT_ISOMETRY(pose_);
-  Eigen::Matrix3d basis = pose_.linear();
-  normalL_ = basis.col(0);
-  normalW_ = basis.col(1);
-  normalH_ = basis.col(2);
+  invRot_ = pose_.linear().transpose();
 
-  const Eigen::Vector3d tmp(normalL_ * length2_ + normalW_ * width2_ + normalH_ * height2_);
+  // rotation is intentionally not applied, the corners are used in intersectsRay()
+  const Eigen::Vector3d tmp(length2_, width2_, height2_);
   corner1_ = center_ - tmp;
   corner2_ = center_ + tmp;
 }
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -666,6 +666,11 @@ bool bodies::Box::intersectsRay(const Eigen::Vector3d& origin, const Eigen::Vect
   const Eigen::Vector3d o(invRot_ * (origin - center_) + center_);
   const Eigen::Vector3d d(invRot_ * dirNorm);
 
+  // tmin and tmax are such values of t in "p = o + t * d" in which the line intersects the box faces.
+  // The box is viewed projected from all three directions, values of t are computed for each of the projections,
+  // and a final constraint on tmin and tmax is updated by each of these projections. If tmin > tmax, there is no
+  // intersection between the line and the box.
+
   divx = 1 / d.x();
   if (divx >= 0)
   {
@@ -724,6 +729,8 @@ bool bodies::Box::intersectsRay(const Eigen::Vector3d& origin, const Eigen::Vect
   if (tzmax < tmax)
     tmax = tzmax;
 
+  // As we're doing intersections with a ray and not a line, cases where tmax is negative mean that the intersection is
+  // with the opposite ray and not the one we are working with.
   if (tmax < 0)
     return false;
 
@@ -731,19 +738,26 @@ bool bodies::Box::intersectsRay(const Eigen::Vector3d& origin, const Eigen::Vect
   {
     if (tmax - tmin > detail::ZERO)
     {
+      // tmax > tmin, we have two distinct intersection points
       if (tmin > detail::ZERO)
       {
+        // tmin > 0, both intersections lie on the ray
         intersections->push_back(tmin * dirNorm + origin);
         if (count == 0 || count > 1)
           intersections->push_back(tmax * dirNorm + origin);
       }
       else
       {
+        // tmin <= 0 && tmax >= 0, the first intersection point is on the opposite ray and the second one on the correct
+        // ray - this means origin of the ray lies inside the box and we should only report one intersection.
         intersections->push_back(tmax * dirNorm + origin);
       }
     }
     else
+    {
+      // tmax == tmin, there is exactly one intersection at a corner or edge
       intersections->push_back(tmax * dirNorm + origin);
+    }
   }
 
   return true;
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -1193,30 +1193,32 @@ bool bodies::ConvexMesh::intersectsRay(const Eigen::Vector3d& origin, const Eige
 
   // transform the ray into the coordinate frame of the mesh
   Eigen::Vector3d orig(i_pose_ * origin);
-  Eigen::Vector3d dr(i_pose_ * dirNorm);
+  Eigen::Vector3d dr(i_pose_.linear() * dirNorm);
 
   std::vector<detail::intersc> ipts;
 
   bool result = false;
 
   // for each triangle
-  const unsigned int nt = mesh_data_->triangles_.size() / 3;
-  for (unsigned int i = 0; i < nt; ++i)
+  const auto nt = mesh_data_->triangles_.size() / 3;
+  for (size_t i = 0; i < nt; ++i)
   {
     Eigen::Vector3d vec(mesh_data_->planes_[mesh_data_->plane_for_triangle_[i]].x(),
                         mesh_data_->planes_[mesh_data_->plane_for_triangle_[i]].y(),
                         mesh_data_->planes_[mesh_data_->plane_for_triangle_[i]].z());
 
-    double tmp = vec.dot(dr);
+    const double tmp = vec.dot(dr);
     if (fabs(tmp) > detail::ZERO)
     {
-      double t = -(vec.dot(orig) + mesh_data_->planes_[mesh_data_->plane_for_triangle_[i]].w()) / tmp;
+      // planes_[...].w() corresponds to the unscaled mesh, so we need to compute it ourselves
+      const double w_scaled_padded = vec.dot(scaled_vertices_->at(mesh_data_->triangles_[3 * i]));
+      const double t = -(vec.dot(orig) + w_scaled_padded) / tmp;
       if (t > 0.0)
       {
-        const int i3 = 3 * i;
-        const int v1 = mesh_data_->triangles_[i3 + 0];
-        const int v2 = mesh_data_->triangles_[i3 + 1];
-        const int v3 = mesh_data_->triangles_[i3 + 2];
+        const auto i3 = 3 * i;
+        const auto v1 = mesh_data_->triangles_[i3 + 0];
+        const auto v2 = mesh_data_->triangles_[i3 + 1];
+        const auto v3 = mesh_data_->triangles_[i3 + 2];
 
         const Eigen::Vector3d& a = scaled_vertices_->at(v1);
         const Eigen::Vector3d& b = scaled_vertices_->at(v2);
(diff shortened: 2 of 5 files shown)


peci1:
This would deserve a few tests, they're coming tomorrow. For the time being I verified the algorithm in RViz.

src/bodies.cpp
```diff
@@ -595,34 +595,40 @@ void bodies::Box::computeBoundingCylinder(BoundingCylinder& cylinder) const
   cylinder.radius = sqrt(a * a + b * b);
 }
 
+// for some reason, the O2 optimization screws up this function which then gives wrong results
```
rhaschke: What?

src/bodies.cpp
```diff
@@ -535,7 +535,7 @@ void bodies::Box::updateInternalData()
   normalW_ = basis.col(1);
   normalH_ = basis.col(2);
 
-  const Eigen::Vector3d tmp(normalL_ * length2_ + normalW_ * width2_ + normalH_ * height2_);
+  const Eigen::Vector3d tmp(length2_, width2_, height2_);
```
rhaschke, line 609: Obviously this changes the semantics of `corner1_` and `corner2_`. Could you please document this in the header? Does this change have implications for other pieces of the code?

Could we remove the helper variables normalL, normalW, normalH?
These are only used in Box::containsPoint(), which could be simplified to:
```
const Eigen::Vector3d aligned = (pose_.rotation().transpose() * (p - center_)).abs();
return aligned[0] <= length2_ && aligned[1] <= width2_ && aligned[2] <= height2_;
```

rhaschke (dismissed):
Generally, this looks good to me.
Please provide a description for this PR, mentioning that this depends on other PRs.
Please also provide the promised unit tests (one of them should be failing before this fix).

src/bodies.cpp
peci1: No idea... That's what I observed. Turning off O2 optimization solved the issue.

src/bodies.cpp
peci1: Ok, I'll have a look at this.

peci1:
I added PR description and separated this PR from the others as it doesn't depend on them semantically. I'll have a look at the tests.

src/bodies.cpp
peci1: I think you definitely don't want to call `rotation()` in `containsPoint` as this function calls SVD if I'm not wrong. The question is if calling `linear()` wouldn't do the same job here (which I think would).

Changing the semantics of `corner1_` and `corner2_` wouldn't change anything in this library except for the ray-box test. The variables are protected, so I think chances that somebody else would be using them is low. But maybe it'd be better to rename them so that people who were using these variables will get a build failure. Or I can leave the old variables there and add new ones, but as the old ones were semantically wrong, I'm more for the option of getting rid of them.

rhaschke: > I think you definitely don't want to call rotation() in containsPoint as this function calls SVD if I'm not wrong.

It was calling SVD, when we used Affine3d, but since we use Isometry3d, Eigen knows thatit can just return the 3x3 rotation sub-matrix. No overhead at all.

> Changing the semantics of corner1_ and corner2_ wouldn't change anything in this library.

I just wanted to be sure. No need to change the variable names. Just document their semantics in the header (there was nothing before).

peci1:
I had a look at this again, rebased on current melodic-devel, and added a lot of tests. First, I only added box-ray tests, but then I decided to add tests for all bodies. And along the way, I've discovered and fixed a lot of other bugs in various more or less related functions, like `containsPoint()` and `samplePoint()`. Check out the updated title and description of this PR.

I would've sent these other fixes as another PRs, but the tests in this one would fail without them. But if it's desired, I can take the effort and split this into more PRs - I just do not want to take this effort with the tests, so I'd separate them as the "final" PR, and the prior ones would come without tests.

I did some API/ABI breaks here. 

1. First, as requested by @rhaschke , I removed the unnecessary internal fields `normalL_`, `normalW_` and `normalH_` of `bodies::Box`. I instead added `invPose_`.
1. When I was already breaking API/ABI, I also changed `MeshData` by adding one more field.
1. Not really API/ABI break, but I changed the meaning of `corner1_` and `corner2_` protected members of `bodies::Box`. It should not affect anything but potential user code that would access these fields. But as they had no docs before, I think the chance there is some other code using them is pretty minimal.

If API/ABI breaks are not desired for melodic-devel, 1) can be avoided completely (and the API break be left to noetic), and 2) can also be worked around somehow (last resort would be the translation-unit-private global variable ugly trick).

For some reason, CI did not trigger here (maybe because of the pending requested changes?), but here's mine which shows okay: https://travis-ci.com/github/peci1/geometric_shapes . I've also run a local test with this PR and the whole Moveit repo. It ran fine, except for a few `ushort_test/MeshFilterTestUnsignedShort.unsigned_short` tests, which however fail for me even on melodic-devel branch of geometric_shapes.

As with my other PRs, this may of course break user code that counted with the buggy behavior. But (thanks to the tests), I'm pretty sure that if the library is used correctly, it should not bring any new problems.

src/bodies.cpp
```diff
@@ -454,9 +454,9 @@ bool bodies::Cylinder::intersectsRay(const Eigen::Vector3d& origin, const Eigen:
     double b = 2.0 * ROD.dot(VD);
     double c = ROD.squaredNorm() - radius2_;
     double d = b * b - 4.0 * a * c;
-    if (d > 0.0 && fabs(a) > detail::ZERO)
+    if (d >= 0.0 && fabs(a) > detail::ZERO)
     {
-      d = sqrt(d);
+      d = (d != 0.0 ? sqrt(d) : 0.0);
```
v4hn: Why would you special-case 0?

v4hn: In another pull-request you started using `Eigen::numexpt::sqrt`.
Are there reasons to stick to one of them consistantly?

If we would need the `#ifdef` for the eigen version everywhere, that would be a reason to leave it out or add it through auxiliary headers...

src/bodies.cpp
```diff
@@ -634,34 +633,47 @@ void bodies::Box::computeBoundingBox(bodies::AABB& bbox) const
   bbox.extendWithTransformedBox(getPose(), 2 * Eigen::Vector3d(length2_, width2_, height2_));
 }
 
+// for some reason, the O2 optimization screws up this function which then gives wrong results
```
v4hn: You removed the actual `GCC` directives again without further comments, but left the comment saying it's broken.

Please either (1) find the gcc bug (report) and link it here, (2) remove the comment if it's not a problem, (3) add the compiler version you used and reference a concrete test that fails.

src/bodies.cpp
```diff
@@ -699,9 +711,16 @@ bool bodies::Box::intersectsRay(const Eigen::Vector3d& origin, const Eigen::Vect
...
-      if (count > 1)
+      if (tmin > detail::ZERO)
+      {
+        intersections->push_back(tmin * dir + origin);
+        if (count > 1)
+          intersections->push_back(tmax * dir + origin);
+      }
+      else
+      {
         intersections->push_back(tmax * dir + origin);
```
v4hn: Does `tmin <= detail::ZERO` really imply origin is inside the box?
If it does, please add a comment to make this clear here.

v4hn (approved):
Please address the inline comments. Aside from them, I approve.

v4hn:
Lastly, this patch conflicted with one of the others I merged. Could you please rebase?

src/bodies.cpp
peci1: You're right, the statement is weird and probably functionally equivalent to what was there before. I'll revert this line.

peci1: Regarding `Eigen::numext::sqrt`: this is from its docs:

> It is essentially equivalent to `using std::sqrt; return sqrt(x);`,  but slightly faster for float/double and some compilers (e.g., gcc), thanks to specializations when SSE is enabled.

So I just thought the compiler could get a chance to use some faster instructions in some cases. But I agree that the ifdefs needed for kinetic would really uglify the code

src/bodies.cpp
peci1: I was unable to reproduce the bug in recent tests. So it probably "got resolved". I'll delete the comment.

src/bodies.cpp
v4hn: Personally, I'm not fully convinced the vectorized sqrt actually make a noticeable difference.
As always, if going for a more unusual construct, benchmark it's actually faster.
Please leave it out for now and maybe later provide a patch to change all calls if benchmarks show it's actually worthwhile.

src/bodies.cpp
peci1: That's it. I added comments to the code which should explain everything.

peci1:
I've made the requested changes.

src/bodies.cpp
v4hn: wonderful, thanks! :sun_with_face:

peci1:
Thanks for the reviews, @v4hn ;)

v4hn:
On Fri, Apr 24, 2020 at 10:17:21AM -0700, Martin Pecka wrote:
> Thanks for the reviews, @v4hn ;)

Thanks for your continuous effort with these patches!
I still fail to understand how all these issues could ever have made it through the years...

peci1:
I think insisting on good tests for PRs is the way to prevent most of such errors for the future. I've seen moveit repos already report some kind of test coverage... What about this repo? Is it planned?

v4hn:
https://github.com/ros-planning/geometric_shapes/pull/120 :-)

include/geometric_shapes/bodies.h
```diff
@@ -384,12 +384,10 @@ class Box : public Body
...
   // pose/padding/scaling-dependent values & values computed for convenience and fast upcoming computations
   Eigen::Vector3d center_;
-  Eigen::Vector3d normalL_;
-  Eigen::Vector3d normalW_;
-  Eigen::Vector3d normalH_;
+  Eigen::Matrix3d invRot_;
 
-  Eigen::Vector3d corner1_;
-  Eigen::Vector3d corner2_;
+  Eigen::Vector3d corner1_;  //!< The translated, but not rotated min corner
```
rhaschke, line 503: What about renaming these vectors to min_corner_ and max_corner_?
As you renamed `normal[L,W,H]` anyway, ABI and API is broken anyway and we need to target this to Noetic only.

include/geometric_shapes/bodies.h
```diff
@@ -458,7 +456,12 @@ class ConvexMesh : public Body
   /** \brief (Used mainly for debugging) Count the number of vertices behind a plane*/
   unsigned int countVerticesBehindPlane(const Eigen::Vector4f& planeNormal) const;
 
-  /** \brief Check if a point is inside a set of planes that make up a convex mesh*/
+  /** \brief Check if the point is inside all halfspaces this mesh consists of (mesh_data_->planes_).
+   *
+   * \note The point is expected to have pose_ "cancelled" (have inverse pose of this mesh applied to it)
+   *       and it is expected to be scaled and padded.
```
rhaschke: What does it mean for a point to be scaled and padded?

src/bodies.cpp
```diff
@@ -497,8 +497,18 @@ bool bodies::Cylinder::intersectsRay(const Eigen::Vector3d& origin, const Eigen:
...
-    intersections->push_back(ipts[i].pt);
+
+  // If a ray hits exactly the boundary between a side and a base, it is reported twice.
+  // We want to only return the intersection once; the loop makes use of the fact that ipts is
+  // already sorted by distance, so if some points are duplicate, they are examined successively.
+  for (const auto& p : ipts)
+  {
+    if (intersections->size() == n)
+      break;
+    if (!intersections->empty() && (p.pt - intersections->back()).squaredNorm() < pow(detail::ZERO, 2))
```
rhaschke: What about using Eigen's `isApprox()` function here?

src/bodies.cpp
```diff
@@ -513,17 +523,8 @@ bool bodies::Box::samplePointInside(random_numbers::RandomNumberGenerator& rng,
...
-  if (fabs(pL) > length2_)
-    return false;
-
-  double pW = v.dot(normalW_);
-  if (fabs(pW) > width2_)
-    return false;
-
-  double pH = v.dot(normalH_);
-  return fabs(pH) <= height2_;
+  const Eigen::Vector3d aligned = (invRot_ * (p - center_)).cwiseAbs();
```
rhaschke: Great simplification! Much more readible! :+1:

src/bodies.cpp
```diff
@@ -640,28 +639,45 @@ bool bodies::Box::intersectsRay(const Eigen::Vector3d& origin, const Eigen::Vect
...
+  // rotate the origin and dir instead. corner1_ and corner2_ are corners with canceled rotation.
+  const Eigen::Vector3d o(invRot_ * (origin - center_) + center_);
+  const Eigen::Vector3d d(invRot_ * dir);
+
+  // tmin and tmax are such values of t in "p = o + t * d" in which the line intersects the box faces.
+  // The box is viewed projected from all three directions, values of t are computed for each of the projections,
+  // and a final constraint on tmin and tmax is updated by each of these projections. If tmin > tmax, there is no
+  // intersection between the line and the box.
+
+  divx = 1 / d.x();
```
rhaschke: The following code seems to be repeating the same pattern for x,y, and z.
Wouldn't it make sense to use a for loop from 0..2 instead and store the results min/max[x,y,z] in vectors?
Probably it's even possible to vectorize the whole operation?

src/bodies.cpp
```diff
@@ -716,8 +748,8 @@ bool bodies::ConvexMesh::containsPoint(const Eigen::Vector3d& p, bool /* verbose
     return false;
   if (bounding_box_.containsPoint(p))
   {
+    // Transform the point to the "base space" of this mesh
     Eigen::Vector3d ip(i_pose_ * p);
-    ip = mesh_data_->mesh_center_ + (ip - mesh_data_->mesh_center_) / scale_;
```
rhaschke, line 741: What a weird code!?!

src/bodies.cpp
```diff
@@ -1002,7 +1035,7 @@ void bodies::ConvexMesh::updateInternalData()
       new shapes::Box(mesh_data_->box_size_.x(), mesh_data_->box_size_.y(), mesh_data_->box_size_.z()));
   bounding_box_.setDimensions(box_shape.get());
```
rhaschke: Use the set*Dirty() functions here?

src/bodies.cpp
```diff
@@ -1094,7 +1127,13 @@ bool bodies::ConvexMesh::isPointInsidePlanes(const Eigen::Vector3d& point) const
...
     const Eigen::Vector4d& plane = mesh_data_->planes_[i];
     Eigen::Vector3d plane_vec(plane.x(), plane.y(), plane.z());
-    double dist = plane_vec.dot(point) + plane.w() - padding_ - 1e-6;
+    // w() needs to be recomputed from a scaled vertex as normally it refers to the unscaled plane
+    // we also cannot simply subtract padding_ from it, because padding of the points on the plane causes a different
+    // effect than adding padding along this plane's normal (padding effect is direction-dependent)
+    const auto scaled_point_on_plane =
+        scaled_vertices_->at(mesh_data_->triangles_[3 * mesh_data_->triangle_for_plane_[i]]);
+    const double w_scaled_padded = -plane_vec.dot(scaled_point_on_plane);
+    const double dist = plane_vec.dot(point) + w_scaled_padded - 1e-6;
```
rhaschke: Can you replace the hard-coded 1e-6 by detail::ZERO or should it be significantly larger?

src/bodies.cpp
```diff
@@ -1142,30 +1181,32 @@ bool bodies::ConvexMesh::intersectsRay(const Eigen::Vector3d& origin, const Eige
 
   // transform the ray into the coordinate frame of the mesh
   Eigen::Vector3d orig(i_pose_ * origin);
-  Eigen::Vector3d dr(i_pose_ * dir);
+  Eigen::Vector3d dr(i_pose_.linear() * dir);
```
rhaschke: Just another weird bug!

rhaschke (commented):
Dear @peci1,
please adapt multiple consecutive uses of set*() to set*Dirty() after having merged #126. Otherwise, this looks good. Thanks a lot for all your effort on this!

include/geometric_shapes/bodies.h
peci1: Okay, I can rename them for noetic. As I'd still like to see this in melodic, should I split this PR into two and try to make this one compatible with melodic? In my view, this approach would be easier than first merging to noetic and then backporting...

As discussed earlier, the API/ABI breaks can be avoided at the cost of minor uglification of code.

src/bodies.cpp
peci1: I'll have a look at it. The idea was early stopping. But if vectorization could come into play, it might be faster than the early stopping. There will just be more problems handling nans...

include/geometric_shapes/bodies.h
rhaschke: To maintain ABI you would need to keep the axis vectors instead of `invRot_`. More ugly, but ok.
The other addition I have seen is `std::map<unsigned int, unsigned int> triangle_for_plane_`.
Didn't look for what it is actually used. Replacing this by a global variable, I would actually *not* support. Why is it needed now? Just to be faster?

src/bodies.cpp
rhaschke: As far as I have seen, you are computing the same expression, componentwise.
Hence, you could do the nan-test componentwise as well. That's not more complicated, is it?

include/geometric_shapes/bodies.h
peci1: Walking on a thin ice, I guess a `Matrix3d` could take up the same amount of memory as three `Vector3d`s. But that could run in many issues once memory alignment comes into play. So I rather give up this idea.

`invRot` would have to be recomputed every time it is used, but as it is only a transpose, it shouldn't be that bad in melodic. I can also imagine storing the rows/cols of the inverted matrix in the abandoned vectors, but I bet the performance could be even worse than transposing it where needed.

The `triangle_for_plane` is really required for efficiency - otherwise mesh-ray intersection tests would need to re-enumrate over all triangles to find which plane the current one belongs to. The global variable trick is something Gazebo people are willing to accept if it is just "temporary" to get a feature to a released version; the development version then implements the nicer, but ABI-breaking version. An example is here: https://github.com/osrf/gazebo/commit/415be4050e955129956084bcb7bb4c2b436cd943 . It creates a global map, which stores references to created objects (ConvexMesh in this case) and maps them to the structure that is needed to be added. Ugly, but works.

Actually, I'd consider making the whole `MeshData` PIMPL for noetic (i.e. moving it to the .cpp file), so that storing additional data will not cause an ABI break any more. The same could be done for the other bodies, too, but there would be the additional cost of dynamic memory allocation. The ConvexMesh allocates memory anyways, so adding one more unique_ptr should not be a problem.

include/geometric_shapes/bodies.h
peci1: I edited the documentation.

src/bodies.cpp
peci1: Good catch, it looks much better.

src/bodies.cpp
peci1: Tests pass even with detail::ZERO, so I'll use that one.

src/bodies.cpp
peci1: I think I found a very nice vectorization and simplification of the algorithm. The tests pass, so it should be correct.

src/bodies.cpp
peci1: Done.

include/geometric_shapes/bodies.h
peci1: I implemented ABI compatibility using the global variable.

peci1:
@rhaschke I did all the changes you requested. ABI compatibility with melodic is achieved by the very last commit.

Thanks for the deep insights. The code looks much better now.

peci1:
I also restructured the commit history. Now I consider all the commits valuable and further squashing is not needed.

src/bodies.cpp
```diff
@@ -1021,7 +1031,8 @@ void bodies::ConvexMesh::updateInternalData()
 
   shapes::Box box_shape(mesh_data_->box_size_.x(), mesh_data_->box_size_.y(), mesh_data_->box_size_.z());
   bounding_box_.setPoseDirty(pose);
-  bounding_box_.setPaddingDirty(padding_);
+  bounding_box_.setPaddingDirty(
+      padding_);  // the padding effect will most likely be smaller, but this is the worst case
```
rhaschke: Please improve the comment (if it's really needed).
```suggestion
  // the padding effect will most likely be smaller, but this is the worst case
  bounding_box_.setPaddingDirty(padding_);  
```

src/bodies.cpp
```diff
@@ -1239,13 +1258,29 @@ bool bodies::ConvexMesh::intersectsRay(const Eigen::Vector3d& origin, const Eige
...
-    for (unsigned int i = 0; i < n; ++i)
-      intersections->push_back(ipts[i].pt);
+    // If a ray hits exactly the boundary between two triangles, it is reported twice;
+    // We only want return the intersection once; the loop makes use of the fact that ipts is
+    // already sorted by distance, so if some points are duplicate, they are examined successively.
+    for (const auto& p : ipts)
+    {
+      if (intersections->size() == n)
+        break;
+      if (!intersections->empty() && (p.pt - intersections->back()).squaredNorm() < pow(detail::ZERO, 2))
```
rhaschke: candidate for Eigen::isApprox again.

src/bodies.cpp
```diff
@@ -514,8 +520,18 @@ bool bodies::Cylinder::intersectsRay(const Eigen::Vector3d& origin, const Eigen:
...
+  // We want to only return the intersection once; the loop makes use of the fact that ipts is
+  // already sorted by distance, so if some points are duplicate, they are examined successively.
+  for (const auto& p : ipts)
+  {
+    if (intersections->size() == n)
+      break;
+    if (!intersections->empty() && p.pt.isApprox(intersections->back(), detail::ZERO))
+      continue;
+    intersections->push_back(p.pt);
+  }
```
rhaschke: This code snippet is used twice (see above).
What about introducing an inline function `detail::filter_duplicates(ipts, intersections)` for it?

rhaschke (commented):
Some minor remarks left.
As I will not cherry-pick to noetic, but revert the last commit, please remove the comment 
`Do not cherry-pick this commit to noetic.` from it.
Thanks a lot for all this work. Hope, we can merge this later today.

peci1:
@rhaschke I addressed your comments.

rhaschke:
@peci1, thank you very much. All your PRs are a great improvement of this code base!
If I may ask for even more, here are my suggested next steps:
- use of set*Dirty() functions (for Melodic)
- API cleanup (for Noetic)
  - Consider moving non-trivial functions from `.h` to `.cpp`.
Usually, only one- or two-liners deserve inlining.
  - `min_corner_` and `max_corner_` as suggested in https://github.com/ros-planning/geometric_shapes/pull/109#discussion_r415056537
  - pimpl pattern for MeshData as suggested in https://github.com/ros-planning/geometric_shapes/pull/109#discussion_r415099445
- any other improvements you can think of :wink:

rhaschke:
I added a `noetic-devel` branch, just in case you need it ...

PR #114, opened 2019-08-08
Apply clang-tidy fixes

rhaschke:
Apply the most important clang-tidy fixes. I just dropped the naming conventions of MoveIt, because they are not followed in this rep.

### DIFF
--- a/src/body_operations.cpp
+++ b/src/body_operations.cpp
@@ -122,7 +122,7 @@ Body* constructBodyFromMsgHelper(const T& shape_msg, const geometry_msgs::Pose& 
   }
   return nullptr;
 }
-}
+}  // namespace bodies
 
 bodies::Body* bodies::constructBodyFromMsg(const shapes::ShapeMsg& shape_msg, const geometry_msgs::Pose& pose)
 {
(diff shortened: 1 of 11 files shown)


src/body_operations.cpp
```diff
@@ -145,9 +145,9 @@ void bodies::computeBoundingSphere(const std::vector<const bodies::Body*>& bodie
 
   // TODO - expand to all body types
   unsigned int vertex_count = 0;
-  for (unsigned int i = 0; i < bodies.size(); i++)
+  for (auto body : bodies)
```
BryceStevenWilley, line 148: Should fix `auto` in loops.

BryceStevenWilley (approved):
LGTM, just fixing auto in loops. 

However, I don't think this will fix https://github.com/ros-planning/moveit/pull/1607#issuecomment-519664734, if that's what it's intended for. I think that's a missing `override` in shapes.  If not, carry on. :)

rhaschke:
> However, I don't think this will fix ros-planning/moveit#1607 (comment), if that's what it's intended for.

That's true. Due to its urgency, I fixed the mentioned issue directly via #113.

PR #110, opened 2019-06-14
Porting geometric_shapes to ROS2

anasarrak:
Can we create a ros2 branch for that?

PR #118, opened 2019-11-15
Port to ROS 2 dashing (new branch)

henningkayser:
This is a fixup of the previous ROS 2 migration attempts and synced with melodic-devel.
I created the 'new' dashing-devel based on melodic devel, so this PR only lists the migration steps.
When this gets merged, I suggest we remove crystal-devel and close the PRs https://github.com/ros-planning/geometric_shapes/pull/96 and  https://github.com/ros-planning/geometric_shapes/pull/110

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,10 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.5)
 project(geometric_shapes)
 
-add_compile_options(-std=c++11)
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/")
 
@@ -24,47 +27,27 @@ endif()
 find_library(ASSIMP_ABS_LIBRARIES NAMES ${ASSIMP_LIBRARIES} assimp HINTS ${ASSIMP_LIBRARY_DIRS} ${ASSIMP_PC_LIBRARY_DIRS})
 set(ASSIMP_LIBRARIES "${ASSIMP_ABS_LIBRARIES}")
 
+find_package(rclcpp REQUIRED)
 find_package(Boost REQUIRED system filesystem)
-
 find_package(console_bridge REQUIRED)
-
+find_package(eigen3_cmake_module REQUIRED)
 find_package(Eigen3 REQUIRED)
-
 find_package(octomap REQUIRED)
-
-find_package(catkin REQUIRED COMPONENTS
-  eigen_stl_containers
-  random_numbers
-  resource_retriever
-  shape_msgs
-  visualization_msgs
-)
-
-catkin_package(
-  INCLUDE_DIRS
-    include
-  LIBRARIES ${PROJECT_NAME} ${OCTOMAP_LIBRARIES}
-  CATKIN_DEPENDS
-    eigen_stl_containers
-    random_numbers
-    shape_msgs
-    visualization_msgs
-  DEPENDS
-    EIGEN3
-    OCTOMAP
-    console_bridge
-  )
-
+find_package(ament_cmake REQUIRED)
+find_package(eigen_stl_containers REQUIRED)
+find_package(random_numbers REQUIRED)
+find_package(resource_retriever REQUIRED)
+find_package(shape_msgs REQUIRED)
+find_package(visualization_msgs REQUIRED)
+find_package(geometry_msgs REQUIRED)
 find_package(Qhull REQUIRED)
 if (HAVE_QHULL_2011)
   add_definitions(-DGEOMETRIC_SHAPES_HAVE_QHULL_2011)
 endif()
 
 include_directories(include)
-include_directories(${EIGEN3_INCLUDE_DIR} ${Boost_INCLUDE_DIR} ${ASSIMP_INCLUDE_DIRS} ${OCTOMAP_INCLUDE_DIRS})
-include_directories(${catkin_INCLUDE_DIRS} ${console_bridge_INCLUDE_DIRS})
 
-add_library(${PROJECT_NAME}
+add_library(${PROJECT_NAME} SHARED
   src/aabb.cpp
   src/bodies.cpp
   src/body_operations.cpp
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,14 @@
...
+cmake_minimum_required(VERSION 3.5)
 project(geometric_shapes)
 
-add_compile_options(-std=c++11)
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
rhaschke: Compiler options shouldn't be imposed by the build system.

CMakeLists.txt
```diff
@@ -18,53 +25,33 @@ if (NOT ASSIMP_FOUND)
...
 # Both, pkg-config and assimp's cmake-config don't provide an absolute library path.
 # For, pkg-config the path is in ASSIMP_PC_LIBRARY_DIRS, for cmake in ASSIMP_LIBRARY_DIRS.
 find_library(ASSIMP_ABS_LIBRARIES NAMES ${ASSIMP_LIBRARIES} assimp HINTS ${ASSIMP_LIBRARY_DIRS} ${ASSIMP_PC_LIBRARY_DIRS})
 set(ASSIMP_LIBRARIES "${ASSIMP_ABS_LIBRARIES}")
-
 find_package(Boost REQUIRED system filesystem)
-
 find_package(console_bridge REQUIRED)
-
+find_package(eigen3_cmake_module REQUIRED)
```
rhaschke, line 33: Is this really needed?

CMakeLists.txt
```diff
@@ -74,19 +61,52 @@ add_library(${PROJECT_NAME}
...
+    QHULL
+)
 
-target_link_libraries(${PROJECT_NAME} ${ASSIMP_LIBRARIES} ${QHULL_LIBRARIES} ${catkin_LIBRARIES} ${console_bridge_LIBRARIES} ${Boost_LIBRARIES})
-
+ament_export_libraries(${PROJECT_NAME})
+ament_export_include_directories(include)
+ament_export_dependencies(
+	Eigen3
+	eigen3_cmake_module  # export Eigen3 headers
```
rhaschke: See above.

include/geometric_shapes/shape_extents.h
```diff
@@ -35,16 +35,17 @@
 #ifndef GEOMETRIC_SHAPES_SHAPE_EXTENTS_
 #define GEOMETRIC_SHAPES_SHAPE_EXTENTS_
 
-#include <shape_msgs/SolidPrimitive.h>
-#include <shape_msgs/Mesh.h>
+#include "shape_msgs/msg/solid_primitive.hpp"
```
rhaschke: You shouldn't change `<...>` to quotes `"..."`. They have a different meaning!
Quotes should be used for local files only.
This applies to all files.

include/geometric_shapes/solid_primitive_dims.h
```diff
@@ -50,60 +50,61 @@ struct SolidPrimitiveDimCount
...
-                   static_cast<int>(shape_msgs::SolidPrimitive::BOX_Z))) ?
-                 static_cast<int>(shape_msgs::SolidPrimitive::BOX_Y) :
-                 ((static_cast<int>(shape_msgs::SolidPrimitive::BOX_Z) >=
-                       static_cast<int>(shape_msgs::SolidPrimitive::BOX_X) &&
-                   static_cast<int>(shape_msgs::SolidPrimitive::BOX_Z) >=
-                       static_cast<int>(shape_msgs::SolidPrimitive::BOX_Y)) ?
-                      static_cast<int>(shape_msgs::SolidPrimitive::BOX_Z) :
-                      0)) +
-                1
+    value = (static_cast<int>(shape_msgs::msg::SolidPrimitive::BOX_X) >=
```
rhaschke, line 66: What the hell is this doing? Just a comment - not related to this PR.

package.xml
```diff
@@ -1,18 +1,23 @@
-<package>
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>geometric_shapes</name>
-  <version>0.6.1</version>
+  <version>0.6.2</version>
```
rhaschke: The version number shouldn't be increased (yet).

package.xml
```diff
@@ -1,18 +1,23 @@
...
   <description>This package contains generic definitions of geometric shapes and bodies.</description>
 
   <author email="isucan@google.com">Ioan Sucan</author>
   <author email="gjones@willowgarage.edu">Gil Jones</author>
 
   <maintainer email="dave@dav.ee">Dave Coleman</maintainer>
   <maintainer email="130s@2000.jukuin.keio.ac.jp">Isaac I. Y. Saito</maintainer>
+  <maintainer email="vmayoral@acutronicrobotics.com">Víctor Mayoral Vilches</maintainer>
```
rhaschke: Does Victor really become a maintainer of this package?

package.xml
```diff
@@ -25,19 +30,32 @@
   <build_depend>random_numbers</build_depend>
   <build_depend>resource_retriever</build_depend>
   <build_depend>shape_msgs</build_depend>
+  <build_depend>geometry_msgs</build_depend>
   <build_depend>visualization_msgs</build_depend>
+  <build_depend>rclcpp</build_depend>
+
+  <exec_depend>rclcpp</exec_depend>
```
rhaschke: Please merge `<build_depend>` and `<exec_depend>` tags into `<depend>`.

rhaschke (changes requested):
Hi Henning, it would be great if you could cleanup the commit history somewhat and have individual commits for:
- msg related changes
- cmake/catkin/ament related changes
- real code changes (probably none)

rhaschke:
> When this gets merged, I suggest we remove crystal-devel and close the PRs #96 and #110.

Fully agree.

CMakeLists.txt
henningkayser: Unfortunately, yes. Otherwise headers are not found and export doesn't work. See
 https://github.com/ros2/eigen3_cmake_module

include/geometric_shapes/solid_primitive_dims.h
henningkayser: It... attempts to find the highest index value between `BOX_X`, `BOX_Y`, `BOX_Z` and adds +1, ignoring that these are defined constants? I have no idea

package.xml
henningkayser: I don't think so. By now this is just a fixup of the existing branch based on Acutronic's fork. I'll remove all unnecessary changes when cleaning up the history

package.xml
henningkayser: Makes sense. Isn't that something we want for melodic-devel as well?

rhaschke: Sure.

henningkayser:
> Hi Henning, it would be great if you could cleanup the commit history somewhat and have individual commits for:
> 
> * msg related changes
> * cmake/catkin/ament related changes
> * real code changes (probably none)

@rhaschke I squashed all commits into a more readable history and applied your requested changes. Travis fails because of warnings in octomap and because there is no CMAKE_BUILD_TYPE defined. For the latter we need to pass the parameter to `colcon build` which requires changes in `moveit_ci`. Since there is nothing to add for this PR, lgtm.

rhaschke:
Rebased onto latest HEAD in #122.

PR #119, opened 2019-11-15
simplify checking of shape dimensions

rhaschke:
Address https://github.com/ros-planning/geometric_shapes/pull/118#discussion_r346757498 and simplify this horrible code.

### DIFF
--- a/include/geometric_shapes/solid_primitive_dims.h
+++ b/include/geometric_shapes/solid_primitive_dims.h
@@ -32,8 +32,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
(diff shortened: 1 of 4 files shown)


henningkayser (commented):
Please fix CI (add missing includes of  `solid_primitive_dims.h`). Then +1 from me

bmagyar (approved):
I'm wondering if with C++14 it is possible to use constexpr instead of the enum trick but it's equally good!

rhaschke:
> I'm wondering if with C++14 it is possible to use constexpr instead of the enum trick

I thought about this as well. But I didn't want to change API for Melodic that much.
Thus, I filed #121 for ROS2.

PR #121, opened 2019-11-17
struct SolidPrimitiveDimCount -> function solidPrimitiveDimCount()

rhaschke:
As suggested in https://github.com/ros-planning/geometric_shapes/pull/119#pullrequestreview-317913368, further simplify `SolidPrimitiveDimCount` by introducing a `constexpr` function.


mamoll (approved):
lgtm

PR #122, opened 2019-11-17
Port to ROS2

rhaschke:
Rebase of #118.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,10 @@ env:
   global:
     - ROS_DISTRO=dashing
     - ROS_REPO=ros
+    - UPSTREAM_WORKSPACE=geometric_shapes.repos
     - CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unused-parameter"
     - WARNINGS_OK=false
+    - BEFORE_DOCKER_SCRIPT="sed -i 's/travis_run --retry rosdep install/#rosdep install/g' .moveit_ci/travis.sh"
   matrix:
     - TEST="clang-format, ament_lint"
     - ROS_DISTRO=dashing
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,23 +13,21 @@ notifications:
       - 130s@2000.jukuin.keio.ac.jp
 env:
   global:
-    - ROS_DISTRO=melodic
+    - ROS_DISTRO=dashing
     - ROS_REPO=ros
     - CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unused-parameter"
     - WARNINGS_OK=false
   matrix:
-    - TEST="clang-format, catkin_lint"
-    - ROS_DISTRO=kinetic
-    - ROS_DISTRO=melodic
+    - TEST="clang-format, ament_lint"
+    - ROS_DISTRO=dashing
 
 matrix:
   include:
     - compiler: clang
       env: TEST=clang-tidy-fix
-           CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unused-parameter"
 
 before_script:
-  - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci
+  - git clone -q -b ros2 --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci
 
 script:
   - .moveit_ci/travis.sh
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,10 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.5)
 project(geometric_shapes)
 
-add_compile_options(-std=c++11)
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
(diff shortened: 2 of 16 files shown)


CMakeLists.txt
```diff
@@ -8,11 +8,6 @@ endif()
 
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/")
 
-if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
```
rhaschke: I suggest to remove this: The cmake file shouldn't impose any (non-required) build setting.

geometric_shapes.repos
```diff
@@ -3,13 +3,7 @@ repositories:
     type: git
     url: https://github.com/ros-planning/geometric_shapes
     version: dashing-devel
-# TODO(henningkayser): use upstream package once https://github.com/ros-planning/random_numbers/pull/18 is merged
```
rhaschke: Done.

geometric_shapes.repos
```diff
@@ -3,13 +3,7 @@ repositories:
...
   random_numbers:
     type: git
-    url: https://github.com/PickNikRobotics/random_numbers
-    version: ros2-devel
-# TODO(henningkayser): merge dependencies below into upstream packages
-  octomap:
```
rhaschke: Why do you need a private version of octomap?
Using this private version, Travis raises a lot of issues, which it doesn't do with the default octomap package.

.travis.yml
```diff
@@ -21,12 +21,12 @@ env:
   matrix:
     - TEST="clang-format, ament_lint"
     - ROS_DISTRO=dashing
+      BEFORE_SCRIPT=apt install liboctomap-dev
```
rhaschke: We need to add octomap to rosdep as a system dependency...

rhaschke (commented):
Before merging this, I suggest to release random_numbers and declare octomap as a system dependency. Thus we can drop the corresponding workarounds for Travis.
@henningkayser, see comments below for my additional changes.

vmayoral:
@henningkayser and @rhaschke, you're self-attributing yourselves commits in here. I don't remember you contributed to this in February. See https://github.com/ros-planning/geometric_shapes/pull/96/commits. It'd be nice if you could rebase and add original authors.

rhaschke:
@vmayoral, Henning has put a lot of effort to cleanup the commits in #96, which were changing stuff back and forth. Probably that's the reason why he put himself as author for cmake related changes.
As you might have noticed, I'm not mentioned as author of any commits (except fixups, which I expect to be squashed with their base commits).

rhaschke:
@henningkayser, I have asked for a release of octomap into dashing.

CMakeLists.txt
v4hn: This check was added throughout MoveIt packages to avoid bad first-time experiences for new users building from source. Solving planning problems can be apparently an order of magnitude slower when building with a standard pipeline and cmake's default.

If this is still the case (I don't see why the situation would have changed) I would vote to keep it.

rhaschke: Ok.

henningkayser:
@rhaschke thanks for following up on the `random_numbers` and `octomap` dependencies! With `octomap` I had some issues with the ros2 branch, namely conflicting opengl headers and issues with the top-level CMakeLists `octomap-distribution` which lead to problems with colcon finding `octomap` (pretty much the same issue you just ran into with Travis CI). `octomap-distribution` is detected as the top-level package by `colcon`. Even with `octomap` included as build dependency, `colcon` doesn't find it and might even build `geometric_shapes` before `octomap-distribution` resulting in missing dependency `octomap`. If we add `octomap-distribution` as build dependency to the `package.xml`, rosdep complains that this package can't be found since it doesn't provide a `package.xml`. Either `octomap-distribution` needs to be a proper meta package with separate `package.xml` or it should be omitted in the ROS 2 package. I wanted to address those issues next, the forked branch contains temporary fixes.

> @henningkayser, I have asked for a release of octomap into dashing.

That's great, with pre-built releases the issues above shouldn't occur

geometric_shapes.repos
henningkayser: See comment below

.travis.yml
henningkayser: Of course, if it is available

geometric_shapes.repos
henningkayser: Thanks!

geometric_shapes.repos
```diff
@@ -0,0 +1,13 @@
...
+    url: https://github.com/ros-planning/geometric_shapes
+    version: dashing-devel
+  random_numbers:
+    type: git
+    url: https://github.com/ros-planning/random_numbers
+    version: ros2
+  octomap:
+    type: git
+    url: https://github.com/ros-gbp/octomap-release.git
+    version: debian/ros-melodic-octomap_1.9.0-1_bionic
```
henningkayser, line 13: I didn't think this would work. Great, if it does!

henningkayser (approved):
Ok, CI is successful now. Can you squash the fixup commits? After that, lgtm

CMakeLists.txt
henningkayser: The only problem I see is that `colcon` prints all messages as warnings (actually, printing to `stderr`) which are treated as errors by Travis. We have to add
`-DCMAKE_BUILD_TYPE=Release` to each `colcon build` command

rhaschke: Done in https://github.com/ros-planning/moveit_ci/pull/95

PR #124, opened 2019-12-30
Resolve octomap dependency with rosdep

henningkayser:
As this should also work for Eloquent, should we rename `dashing-devel` to `ros2`?


mlautman (approved):

rhaschke (approved):

rhaschke:
> As this should also work for Eloquent, should we rename dashing-devel to ros2?

Definitely.

Issue #125, opened 2020-03-13
Performance: Body::setPose() should use Eigen::Isometry3d::linear() instead of ::rotation().


peci1:
Here: https://github.com/ros-planning/geometric_shapes/blob/bca6b88be8ba70851c79f785e0d45b12cb1917df/src/bodies.cpp#L318

And here: https://github.com/ros-planning/geometric_shapes/blob/bca6b88be8ba70851c79f785e0d45b12cb1917df/src/bodies.cpp#L549

When using rotation(), Eigen unnecessarily does SVD decomposition via `Eigen::Transform::computeRotationScaliing()`. But we're working with isometries, so linear part of the transform is equal to the rotation. I've done some performance tests and calling `setPose()` on 20 000 bodies takes 250 ms, which is a lot (and according to the profiler, the time is dominated by SVD computations).

peci1:
A guy on SO says [newer Eigen should recognize and use the potential optimization](https://stackoverflow.com/a/58680648/1076564), but at least the version used in Melodic doesn't do that.

v4hn:
Could you create a pull-request and in the best case benchmark the performance improvement?
If the `Isometry` template indeed does not optimize away the SVD, this impacts MoveIt in general just as well.

peci1:
I'll try that. Just as a confirmation, here's the disassembly of /opt/ros/melodic/lib/libgeometric_shapes.so version 0.6.1-0bionic.20191205.163901 (created by `objdump -d`):

    0000000000014370 <_ZN6bodies8Cylinder18updateInternalDataEv@@Base>:
        ...
        1442e:	e8 dd 82 ff ff       	callq  c710 <_ZNK5Eigen9TransformIdLi3ELi1ELi0EE22computeRotationScalingINS_6MatrixIdLi3ELi3ELi0ELi3ELi3EEES4_EEvPT_PT0_@plt>

The docs for `Transform::rotation()` in Eigen master say:

> If Mode==Isometry, then this method is an alias for linear(), otherwise it calls computeRotationScaling() to extract the rotation through a SVD decomposition.

But apparently, `computeRotationAndScaling` gets called.

Eigen development branch 3.3 doesn't contain this optimization: https://gitlab.com/libeigen/eigen/-/blob/3.3/Eigen/src/Geometry/Transform.h#L1061 .
[...]

PR #126, opened 2020-03-22
Improve performance of bodies::setPose()

peci1:
Performance of bodies::setPose() was improved by calling .linear() instead of .rotation().
For isometries, these should result to the same code, but this optimization is not in any released version of Eigen so far (3.3.7).

I also had to fix test_bounding_box.cpp test, as I found there a few usages of AngleAxis to which a denormalized axis was passed, which is illegal. And the fixes in this PR broke one of the wrong tests.

I also did additional speedups in ConvexMesh::setPose(). 
First, there was an unncessary dynamic memory allocation, which was not optimized out by the compiler (removing the allocation speeded up the pref. test by ~25%).
Second, the construction of the helper bounding_box_ object led to calling Box::updateInternalData() 4 times in succession. So I fixed that by accessing the protected members of Box and making ConvexMesh its friend class. If this approach is not desirable, I extracted it to a single commit which can be reverted.

For the measured performance improvement, see https://github.com/ros-planning/geometric_shapes/issues/125#issuecomment-602134891.

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -71,9 +71,35 @@ struct BoundingCylinder
   Eigen::Isometry3d pose;
   double radius;
   double length;
-
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 };
+
+// To be able to use the more efficient Eigen::Transform::linear() instead of rotation(),
+// we need to check the user has really passed an isometry. To avoid runtime costs,
+// this check is only done as assert, which get compiled-out in release builds
+#ifdef NDEBUG
+#define ASSERT_ISOMETRY(transform) assert(true);
+#else
+inline void checkIsometry(const Eigen::Isometry3d& transform)
+{
+  if (!transform.matrix().row(3).isApprox(Eigen::Vector4d::UnitW().transpose()))
+  {
+    std::cerr << "The given transform is not an isometry! It's last row is: " << std::endl
+              << transform.matrix().row(3) << std::endl;
+    assert(!"Invalid isometry transform");
+  }
+
+  Eigen::Isometry3d::LinearMatrixType scale;
+  transform.computeRotationScaling((Eigen::Isometry3d::LinearMatrixType*)nullptr, &scale);
+  if (!scale.isApprox(Eigen::Matrix3d::Identity()))
+  {
+    std::cerr << "The given transform is not an isometry! It's linear part involves scaling: " << std::endl
+              << scale.diagonal().transpose() << std::endl;
+    assert(!"Invalid isometry transform");
+  }
+}
+#define ASSERT_ISOMETRY(transform) ::bodies::checkIsometry(transform);
+#endif
 
 class Body;
 
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -104,11 +104,24 @@ public:
     return type_;
   }
 
+  /**
+   * \brief If the dimension of the body should be scaled, this method sets the scale.
+   * \note This is the dirty version of the function which does not update internal data that depend on the scale.
+   *       In the general case, you should call setScale() instead. Only call this function if you have a series of
+   *       calls like setScale/setPadding/setPose/setDimensions and you want to avoid the overhead of updating the
+   *       internal structures after each call. When you are finished with the batch, call updateInternalData().
+   * \param scale The scale to set. 1.0 means no scaling.
+   */
+  void setScaleDirty(double scale)
+  {
+    scale_ = scale;
+  }
+
   /** \brief If the dimension of the body should be scaled, this
       method sets the scale. Default is 1.0 */
   void setScale(double scale)
   {
-    scale_ = scale;
+    setScaleDirty(scale);
     updateInternalData();
   }
 
@@ -118,11 +131,24 @@ public:
     return scale_;
   }
 
+  /**
+   * \brief If the dimension of the body should be padded, this method sets the pading.
+   * \note This is the dirty version of the function which does not update internal data that depend on the scale.
+   *       In the general case, you should call setPadding() instead. Only call this function if you have a series of
+   *       calls like setScale/setPadding/setPose/setDimensions and you want to avoid the overhead of updating the
+   *       internal structures after each call. When you are finished with the batch, call updateInternalData().
+   * \param padd The padding to set (in meters). 0.0 means no padding.
+   */
+  void setPaddingDirty(double padd)
+  {
+    padding_ = padd;
+  }
+
   /** \brief If constant padding should be added to the body, this
       method sets the padding. Default is 0.0 */
   void setPadding(double padd)
   {
-    padding_ = padd;
+    setPaddingDirty(padd);
     updateInternalData();
   }
 
@@ -132,10 +158,23 @@ public:
     return padding_;
   }
 
+  /**
+   * \brief Set the pose of the body.
+   * \note This is the dirty version of the function which does not update internal data that depend on the pose.
+   *       In the general case, you should call setPose() instead. Only call this function if you have a series of
+   *       calls like setScale/setPadding/setPose/setDimensions and you want to avoid the overhead of updating the
+   *       internal structures after each call. When you are finished with the batch, call updateInternalData().
+   * \param pose The pose to set. Default is identity.
+   */
+  void setPoseDirty(const Eigen::Isometry3d& pose)
+  {
+    pose_ = pose;
+  }
+
   /** \brief Set the pose of the body. Default is identity */
   void setPose(const Eigen::Isometry3d& pose)
   {
-    pose_ = pose;
+    setPoseDirty(pose);
     updateInternalData();
   }
 
@@ -143,6 +182,19 @@ public:
   const Eigen::Isometry3d& getPose() const
   {
     return pose_;
+  }
+
+  /**
+   * \brief Set the dimensions of the body (from corresponding shape).
+   * \note This is the dirty version of the function which does not update internal data that depend on the dimensions.
+   *       In the general case, you should call setDimensions() instead. Only call this function if you have a series of
+   *       calls like setScale/setPadding/setPose/setDimensions and you want to avoid the overhead of updating the
+   *       internal structures after each call. When you are finished with the batch, call updateInternalData().
+   * \param shape The shape whose dimensions should be assumed. After the function finishes, the pointer can be deleted.
+   */
+  inline void setDimensionsDirty(const shapes::Shape* shape)
+  {
+    useDimensions(shape);
   }
 
   /** \brief Get the dimensions associated to this body (as read from corresponding shape) */
@@ -205,12 +257,12 @@ public:
       thread safety, when bodies need to be moved around. */
   virtual BodyPtr cloneAt(const Eigen::Isometry3d& pose, double padding, double scaling) const = 0;
 
-protected:
   /** \brief This function is called every time a change to the body
       is made, so that intermediate values stored for efficiency
       reasons are kept up to date. */
   virtual void updateInternalData() = 0;
 
+protected:
   /** \brief Depending on the shape, this function copies the relevant data to the body. */
   virtual void useDimensions(const shapes::Shape* shape) = 0;
 
@@ -264,9 +316,10 @@ public:
 
   virtual BodyPtr cloneAt(const Eigen::Isometry3d& pose, double padding, double scale) const;
 
+  virtual void updateInternalData();
+
 protected:
   virtual void useDimensions(const shapes::Shape* shape);
-  virtual void updateInternalData();
 
   // shape-dependent data
   double radius_;
@@ -314,9 +367,10 @@ public:
 
   virtual BodyPtr cloneAt(const Eigen::Isometry3d& pose, double padding, double scale) const;
 
+  virtual void updateInternalData();
+
 protected:
   virtual void useDimensions(const shapes::Shape* shape);
-  virtual void updateInternalData();
 
   // shape-dependent data
   double length_;
@@ -374,9 +428,10 @@ public:
 
   virtual BodyPtr cloneAt(const Eigen::Isometry3d& pose, double padding, double scale) const;
 
+  virtual void updateInternalData();
+
 protected:
   virtual void useDimensions(const shapes::Shape* shape);  // (x, y, z) = (length, width, height)
-  virtual void updateInternalData();
 
   // shape-dependent data
   double length_;
@@ -452,9 +507,10 @@ public:
 
   void correctVertexOrderFromPlanes();
 
+  virtual void updateInternalData();
+
 protected:
   virtual void useDimensions(const shapes::Shape* shape);
-  virtual void updateInternalData();
 
   /** \brief (Used mainly for debugging) Count the number of vertices behind a plane*/
   unsigned int countVerticesBehindPlane(const Eigen::Vector4f& planeNormal) const;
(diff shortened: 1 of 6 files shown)


peci1:
As was suggested in #125, I didn't create a performance test that would be a part of the package, because the rules on how to write them in a non-flaky way are unclear.

Anyways, if anybody wanted to replicate my experiments, here's the test code:

```
/*********************************************************************
* Software License Agreement (BSD License)
*
*  Copyright (c) 2019, Open Robotics.
*  All rights reserved.
*
*  Redistribution and use in source and binary forms, with or without
*  modification, are permitted provided that the following conditions
*  are met:
*
*   * Redistributions of source code must retain the above copyright
*     notice, this list of conditions and the following disclaimer.
*   * Redistributions in binary form must reproduce the above
*     copyright notice, this list of conditions and the following
*     disclaimer in the documentation and/or other materials provided
*     with the distribution.
*   * Neither the name of the Open Robotics nor the names of its
*     contributors may be used to endorse or promote products derived
*     from this software without specific prior written permission.
*
*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
*  POSSIBILITY OF SUCH DAMAGE.
*********************************************************************/

/** \Author Martin Pecka */

#include <geometric_shapes/bodies.h>
#include <gtest/gtest.h>
#include <ros/duration.h>
#include <ros/time.h>
#include <geometric_shapes/mesh_operations.h>
#include "resources/config.h"


TEST(BodiesPerformance, Sphere)
{
  shapes::Sphere shape(1.0);
  bodies::Sphere body(&shape);
  Eigen::Isometry3d tf = Eigen::Isometry3d::Identity();

  const auto start = ros::WallTime::now();
  for (size_t i = 0; i < 10000000; ++i)
    body.setPose(tf);
  const auto end = ros::WallTime::now();

  EXPECT_GT(1.0, (end - start).toSec());
}

TEST(BodiesPerformance, Box)
{
  shapes::Box shape(1.0, 1.0, 1.0);
  bodies::Box body(&shape);
  Eigen::Isometry3d tf = Eigen::Isometry3d::Identity();

  const auto start = ros::WallTime::now();
  for (size_t i = 0; i < 10000000; ++i)
    body.setPose(tf);
  const auto end = ros::WallTime::now();

  EXPECT_GT(1.0, (end - start).toSec());
}

TEST(BodiesPerformance, Cylinder)
{
  shapes::Cylinder shape(1.0, 1.0);
  bodies::Cylinder body(&shape);
  Eigen::Isometry3d tf = Eigen::Isometry3d::Identity();

  const auto start = ros::WallTime::now();
  for (size_t i = 0; i < 10000000; ++i)
    body.setPose(tf);
  const auto end = ros::WallTime::now();

  EXPECT_GT(1.0, (end - start).toSec());
}

TEST(BodiesPerformance, Mesh)
{
  shapes::Mesh* shape = shapes::createMeshFromResource("file://" + std::string(TEST_RESOURCES_DIR) + "/box.dae");
  bodies::ConvexMesh body(shape);
  delete shape;
  Eigen::Isometry3d tf = Eigen::Isometry3d::Identity();

  const auto start = ros::WallTime::now();
  for (size_t i = 0; i < 10000000; ++i)
    body.setPose(tf);
  const auto end = ros::WallTime::now();

  EXPECT_GT(1.0, (end - start).toSec());
}

TEST(BodiesPerformance, ScaledMesh)
{
  shapes::Mesh* shape = shapes::createMeshFromResource("file://" + std::string(TEST_RESOURCES_DIR) + "/box.dae");
  bodies::ConvexMesh body(shape);
  delete shape;
  body.setScale(1.1);
  Eigen::Isometry3d tf = Eigen::Isometry3d::Identity();

  const auto start = ros::WallTime::now();
  for (size_t i = 0; i < 10000000; ++i)
    body.setPose(tf);
  const auto end = ros::WallTime::now();

  EXPECT_GT(1, (end - start).toSec());
}

int main(int argc, char** argv)
{
  testing::InitGoogleTest(&argc, argv);
  return RUN_ALL_TESTS();
}
```

v4hn:
Things to do on a Saturday night. :sunglasses: 
Thank you so much for your analysis and patches!

The `friend` approach to access internal members is indeed not ideal.
What do you think about extending the interface of the `Body` class to include a `setAttributes(pose,padding,scale)` method instead?

peci1:
You're right, the quarantine has some positive effects for open source, I'd say.

I was also thinking about adding a method like the one you suggest, maybe rather with pointer arguments to allow setting just some of them. Also `setDimensions()` is another one that would be good to add. But, eh, having a 4-arg function doesn't seem really nice to me.

Another approach I can think of is adding an optional parameter to `setPose()` etc., like `updateInternalData=true`, which would tell the function if it should call the `updateInternalData()` or not. And leaving a note in the documentation that when the user calls this function with `false` argument, then he is responsible for calling one of these functions with a `true` in the end. Or we could even make `updateInternalData()` public, but that doesn't look good to me. Adding these optional arguments should not affect ABI (as all the relevant functions are non-vritual). I'm just not sure how C++ handles optional arguments API-wise, so it might instead be safer to not alter the existing signatures and adding new 2-arg functions like `setPose(pose, updateInternalData)` (which I think is even recommended by some people over default argument values).

Yet another approach would be to add some "factory-like" object, e.g. `AtrributeSetter`, where you'd set any of these 4 parameters, and then you'd pass the setter object to e.g. `Body::setAttributes(AttributeSetter)`. But I don't personally like this approach very much as it seems overcomplicated to me.

v4hn:
> Another approach I can think of is adding an optional parameter to setPose() etc., like updateInternalData=true, which would tell the function if it should call the updateInternalData() or not. And leaving a note in the documentation that when the user calls this function with false argument, then he is responsible for calling one of these functions with a true in the end. Or we could even make updateInternalData() public, but that doesn't look good to me. Adding these optional arguments should not affect ABI (as all the relevant functions are non-vritual). I'm just not sure how C++ handles optional arguments API-wise, so it might instead be safer to not alter the existing signatures and adding new 2-arg functions like setPose(pose, updateInternalData) (which I think is even recommended by some people over default argument values).

You mixed up API and ABI: a new optional parameter breaks ABI, not API.
For `geometric_shapes` we apparently never adopted `SONAMES`, so an ABI-stable solution is strongly preferred, unless we merge these changes in a new branch (e.g., `master`) only.
I would support adding new interface methods to change the values without updating the datastructures, but adding another parameter does not make sense then.
Instead, I would propose a new family of setters, e.g., `setPoseUnsafe`. This makes it rather obvious people should be careful when to call it :smiley:

peci1:
Okay, adding new setters sounds reasonable. I'm just thinking whether to use `Unsafe` of `Dirty`. But what about `setDimensions()`? This method already has the dirty counterpart, `useDimensions()`, which is protected. Should I create a public version `setDimensionsUnsafe()`, or should I just make `useDimensions()` public? Personally, I'd go with the former option as that would make the naming consistent with the other methods.

v4hn:
> I'm just thinking whether to use Unsafe of Dirty

Either is fine.

> Should I create a public version setDimensionsUnsafe(), or should I just make useDimensions() public?

Consistent naming is important. I prefer `setDimensions*` too.

peci1:
I added the `setScaleDirty()`, `setPaddingDirty()`, `setPoseDirty()` and `setDimensionsDirty()` and made `updateInternalData()` public in all classes. The friend specification was removed.

v4hn (approved):
Looks good to me and does not break API/ABI.

The only unresolved question is whether we want to merge 4f5a7be2a6a08001f1d95964aab471160f6e9983 in melodic-devel still.
Once this is decided (tomorrows meeting), this can be merged as appropriate.

peci1:
I added the isometry asserts. Once we agree on the correct approach, I'll update the moveit PR, too.

Here I added a "no-op" message as the first test in the assert so that when code fails on the assert, the user will see something more friendly than just a bunch of mathematical computations (C++ prints out the whole definition of the assert that failed).

peci1:
The error message looks exactly like this: https://travis-ci.org/github/ros-planning/geometric_shapes/jobs/667849514#L911 :-D

Any idea why the test succeeded on melodic and failed on kinetic?

peci1:
Ah, kinetic build was choosing the integral version of `::abs(int)` instead of `std::abs(double)`. No idea why. I added `std::` to disambiguate.

I also made the assert macro a bit more verbose so that the user can see not only the failed condition, but also the value of the transformation matrix and its computed determinant. This comes at the cost of uglier macro, but provides better user experience.

v4hn (approved):
I very much like the current approach and clean implementation.

Ready for merge in my opinion. I'm still happy to merge this in `melodic-devel`, alternatively we have to split off a `noetic-devel` branch already.

include/geometric_shapes/bodies.h
```diff
@@ -71,10 +71,29 @@ struct BoundingCylinder
...
+// really passed an isometry; to avoid runtime costs, this check is only done as assert, which get compiled-out in
+// release builds
+#ifdef NDEBUG
+#define ASSERT_ISOMETRY(transform) assert(true);
+#else
+#define ASSERT_ISOMETRY(transform)                                                                                     \
+  {                                                                                                                    \
+    if ((transform).matrix()(3, 0) != 0 || (transform).matrix()(3, 1) != 0 || (transform).matrix()(3, 1) != 0 ||       \
+        (transform).matrix()(3, 3) != 1 ||                                                                             \
+        std::abs(1.0 - std::abs((transform).linear().determinant())) >= detail::ZERO)                                  \
```
rhaschke: Actually you should drop the abs() here, because the determinant should be positive.
Otherwise, it's a mirroring operation!

rhaschke: Just checking the determinant isn't sufficient: Let's say the matrix scales by 1.0, 2.0, 0.5 along x,y,z.
The resulting determinant is still 1.0. You actually need to check for equality of the scaling matrix to the identity.

src/bodies.cpp
```diff
@@ -106,9 +106,14 @@ struct interscOrder
 }  // namespace detail
 }  // namespace bodies
 
-void bodies::Body::setDimensions(const shapes::Shape* shape)
+void bodies::Body::setDimensionsDirty(const shapes::Shape* shape)
```
rhaschke: This function should be moved as inline into the header. It just provides another name for useDimensions().
Maybe you can even use [aliasing](https://www.fluentcpp.com/2017/10/27/function-aliases-cpp/)?

rhaschke (commented):
Generally looks good. Some minor remarks though.

include/geometric_shapes/bodies.h
peci1: Okay, I rewrote the test completely, check it out now.

peci1: And I ran test_bounding_box in debug mode locally and it worked. When I passed it a non-isometry, it failed as it should.

src/bodies.cpp
peci1: I used inlining. The aliasing didn't seem to work for me for some reason.

peci1:
@rhaschke This should be ready to merge.

peci1:
There was a forgotten non-isometry in test_bounding_cylinder. It did not trigger locally for me beacause this branch was not up-to-date with melodic-devel. I rebased and now let's hope the tests will pass.

include/geometric_shapes/bodies.h
```diff
@@ -74,30 +74,24 @@ struct BoundingCylinder
...
-// release builds
+// To be able to use the more efficient Eigen::Transform::linear() instead of rotation(),
+// we need to check the user has really passed an isometry. To avoid runtime costs,
+// this check is only done as assert, which get compiled-out in release builds
 #ifdef NDEBUG
 #define ASSERT_ISOMETRY(transform) assert(true);
 #else
-#define ASSERT_ISOMETRY(transform)                                                                                     \
-  {                                                                                                                    \
-    bool isIsometry = (transform).matrix()(3, 0) == 0 && (transform).matrix()(3, 1) == 0 &&                            \
```
peci1: @rhaschke Are you sure this check is done by design? Wouldn't that only be the case in the "efficient storage" which only stores a 3x4 matrix?

test/test_basics.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef NDEBUG
+TEST(Utils, checkIsometry)
+{
+  Eigen::Isometry3d t(Eigen::AngleAxisd(0.42, Eigen::Vector3d(1, 1, 1).normalized()));
+  ASSERT_ISOMETRY(t);
+
+  t.linear() = t.linear() * Eigen::DiagonalMatrix<double, 3, 3>(1.0, 2.0, 3.0);
+  ASSERT_DEATH(ASSERT_ISOMETRY(t), "Invalid isometry transform");
+}
```
peci1: Shouldn't the test do at least something in release mode? I.e. test that calling the assert on invalid transform would not actually do anything?

rhaschke: Yes, I can change this.
Do you want to keep the different commits of this PR or can I squash-merge it?

include/geometric_shapes/bodies.h
rhaschke: True. Will revert.

test/test_basics.cpp
peci1: Well, the last two or three commits do not carry any useful information, but I think the previous ones are nicely "decoupled" and explain why each change was made. I'd vouch for keeping these.

rhaschke: No problem. In this case, I ask you to cleanup the history along the following line:

```
pick 074502d Normalize the direction vector passed to Body::intersectsRay() (#115)
pick e2f4850 Improved performance of Body::setPose() by calling .linear() instead of .rotation().
pick 8dfe31c Improved performance of ConvexMesh::setPose() by getting rid of unnecessary dynamic memory allocation.
pick 13d662e Fixed test_bounding_box.cpp wrong usage of AngleAxis with denormalized axes.
squash e170bcf Fixed a non-isometry in test_bounding_cylinder
pick 5381162 Added dirty versions of setScale/setPadding/setPose/setDimensions and made updateInternalData public.
pick 9c678de Added asserts that check bodies::Body::pose_ is an actual isometry before calling .linear() on it.
fixup 8555637 Fixed ASSERT_ISOMETRY, made setDimensionsDirty() inline.
fixup 721e0b1 Cleanup of ASSERT_ISOMETRY
fixup f3a3af5 fixup! Cleanup of ASSERT_ISOMETRY
```
Obviously, the commit `8555637 Fixed ASSERT_ISOMETRY, made setDimensionsDirty() inline.` needs to be split up onto two previous commits.

peci1:
I cleaned up the commit history.

PR #128, opened 2020-03-26
Added constructors of bodies from the corresponding bounding shapes

peci1:
Added constructors `explicit bodies::Sphere(const bodies::BoundingSphere& sphere)`, and analogously for `Box<->AABB` and `Cylinder<->BoundingCylinder`.

These would help in some tests I'm planning to write.

Two comments:

1. Another option would be to instead add methods e.g. `bodies::BoundingSphere::toBody()`. I just chose the constructor way because I like it more, but they should be more or less equivalent.
1. Why are so many methods in bodies implemented in header files? Shouldn't I move the definitions to .cpp? I just wanted to be consistent, so I've put it in the header as the other constructors, but I don't really see a reason for not putting them into .cpp.

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -244,6 +244,17 @@ public:
     setDimensions(shape);
   }
 
+  explicit Sphere(const BoundingSphere& sphere) : Body()
+  {
+    type_ = shapes::SPHERE;
+    shapes::Sphere shape(sphere.radius);
+    setDimensions(&shape);
+
+    Eigen::Isometry3d pose = Eigen::Isometry3d::Identity();
+    pose.translation() = sphere.center;
+    setPose(pose);
+  }
+
   virtual ~Sphere()
   {
   }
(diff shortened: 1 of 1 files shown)


v4hn (approved):
Makes sense.

> Another option would be to instead add methods e.g. bodies::BoundingSphere::toBody(). I just chose the constructor way because I like it more, but they should be more or less equivalent.

Your approach is fine.

> Why are so many methods in bodies implemented in header files?
> Shouldn't I move the definitions to .cpp?
> I just wanted to be consistent, so I've put it in the header as the other constructors, but I don't really see a reason for not putting them into .cpp.

The reason was probably expected gains from implicit inlining.
I don't know whether there is actually a measurable benefit on modern systems.

These new constructors could also benefit from the proposed `set*Dirty` methods in the future.

rhaschke (changes requested):
Looks good. Please use the set*Dirty() functions as suggested in https://github.com/ros-planning/geometric_shapes/pull/128#pullrequestreview-381848386.

peci1:
This dirty methods depend on #126. I'll integrate them once it is merged.

rhaschke:
> This dirty methods depend on #126. I'll integrate them once it is merged.

My plan is to merge in the following order:
1. #126 
2. #128 
3. #109 

Please address my comments in this order and send me a note if you are done with any of them, so that I can merge and you can proceed with those set*Dirty() methods.

rhaschke:
Hi @peci1, #126 and #128 are already merged now.
Next step is to fix ABI compatibility in #109 and finally, you could file a new PR to make use of the new set*Dirty() functions whereever meaningful. Thanks for your effort and your responsiveness today!

PR #129, opened 2020-04-02
Bump CMake version to avoid CMP0048 warning

ahcorde:
This bumps the minimum CMake version to 3.0.2, which is the [minimum supported by ROS Kinetic](https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021) and new enough to default to the NEW behavior of [CMP0048](https://cmake.org/cmake/help/v3.0/policy/CMP0048.html). This avoids a CMake warning when building and testing this package in Debian Buster and Ubuntu Focal.

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.0.2)
 project(geometric_shapes)
 
 add_compile_options(-std=c++11)


v4hn (approved):
I've lived with this warning for a long time but did not know it was that easy to take care of :+1: 
Though apparently cmake stopped uttering the warning at some point and I don't see it anymore on cmake 3.17 .

v4hn:
followup: I can still reproduce the warning with 3.17 under some circumstances, but it does not appear for every package for me. Either way, bumping the minimal version to 3.0.2 does not harm with indigo out of support.

PR #130, opened 2020-04-10
Fix Eigen Memory alignment due to std::make_shared

gleichdick:
`std::make_shared` does not honor memoy alignment as needed by Eigen objects. So, shared Pointers to classes with `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` must not be created with `std::make_shared()`.

I switched on vectorization with AVX, so the alignment is 32 bytes. The point inclusion test failed sometimes by throwing an Eigen alignment assertation.

### DIFF
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -71,7 +71,7 @@ namespace
 {
 // aligned version of std::make_shared
 template <class T, class... Args>
-std::shared_ptr<T> aligned_make_shared(Args... args)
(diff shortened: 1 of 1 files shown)


v4hn:
We will likely not add any auxiliary alignment code as discussed in [the corresponding pr for moveit](https://github.com/ros-planning/moveit/pull/2011#pullrequestreview-394910619).

At least someone should test the c++17 approach and see whether these (or similar) patches are really still needed there.

gleichdick:
closing this as well

PR #133, opened 2020-04-24
Make bodies::samplePointInside const.

peci1:
It does not need to modify anything in the body.


v4hn:
Merging as trivial.

Thanks.

PR #135, opened 2020-04-27
Added a mutex that protects bodies::detail::g_triangle_for_plane_.

### LINKED ISSUE #134: Add mutex to protect global map
@peci1, please address https://github.com/ros-planning/geometric_shapes/commit/ed4cf1339cf3765ae9ffa6e6fd111a4e342c5fa2#r38767382. Thanks.

peci1:
Fixes #134.

I just hope that taking a pointer of the map's value is okay. The STL docs say:

> References and pointers to either key or data stored in the container are only invalidated by erasing that element, even when the corresponding iterator is invalidated. 

So it should be okay. We're only erasing in ConvexMesh destructor, and at that time, none of the reading functions should be running for this particular mesh.

### DIFF
--- a/src/bodies.cpp
+++ b/src/bodies.cpp
@@ -68,6 +68,7 @@ extern "C" {
 #include <algorithm>
 #include <Eigen/Geometry>
 #include <unordered_map>
+#include <mutex>
 
 namespace bodies
 {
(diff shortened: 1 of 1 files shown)


rhaschke (approved):
Thanks @peci1. I slightly refactored the code to avoid multiple searches in a row.

PR #136, opened 2020-04-28
noetic-abi-breaks

peci1:
API and ABI breaking changes requested in https://github.com/ros-planning/geometric_shapes/pull/109#issuecomment-619560363.


rhaschke:
Applied via #141 (required rebase).

PR #137, opened 2020-04-28
Added a function to create an empty body from shape type.

peci1:
This allows more efficient body construction when scale, padding or pose should also be set during the construction.

This would allow to further optimize https://github.com/ros-planning/moveit/pull/2038 .

### DIFF
--- a/src/body_operations.cpp
+++ b/src/body_operations.cpp
@@ -39,29 +39,40 @@
 #include <console_bridge/console.h>
 #include <Eigen/Geometry>
 
+bodies::Body* bodies::createEmptyBodyFromShapeType(const shapes::ShapeType& shapeType)
+{
+  Body* body = nullptr;
+
+  switch (shapeType)
+  {
+    case shapes::BOX:
+      body = new bodies::Box();
+      break;
+    case shapes::SPHERE:
+      body = new bodies::Sphere();
+      break;
+    case shapes::CYLINDER:
+      body = new bodies::Cylinder();
+      break;
+    case shapes::MESH:
+      body = new bodies::ConvexMesh();
+      break;
+    default:
+      CONSOLE_BRIDGE_logError("Creating body from shape: Unknown shape type %d", (int)shapeType);
+      break;
+  }
+  return body;
+}
+
 bodies::Body* bodies::createBodyFromShape(const shapes::Shape* shape)
 {
   Body* body = nullptr;
 
   if (shape)
-    switch (shape->type)
-    {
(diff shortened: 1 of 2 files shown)


src/body_operations.cpp
```diff
@@ -39,29 +39,40 @@
...
-      case shapes::MESH:
-        body = new bodies::ConvexMesh(shape);
-        break;
-      default:
-        CONSOLE_BRIDGE_logError("Creating body from shape: Unknown shape type %d", (int)shape->type);
-        break;
-    }
+  {
+    body = createEmptyBodyFromShapeType(shape->type);
+    body->setDimensions(shape);
```
rhaschke, line 74: Doesn't that change the semantics of `createBodyFromShape()`?
Before it wasn't calling `setDimensions()`.

src/body_operations.cpp
```diff
@@ -116,7 +127,9 @@ Body* constructBodyFromMsgHelper(const T& shape_msg, const geometry_msgs::Pose&
         q = Eigen::Quaterniond(1.0, 0.0, 0.0, 0.0);
       }
       Eigen::Isometry3d af(Eigen::Translation3d(pose.position.x, pose.position.y, pose.position.z) * q);
-      body->setPose(af);
+      body->setPoseDirty(af);
+      body->setDimensionsDirty(shape);
```
rhaschke, line 131: Same here, `setDimensions()` wasn't called before.

rhaschke (commented):
Before we apply such semantics modifying changes, we should agree with more maintainers.
Can you please elaborate, why these semantics changes are required / useful?

src/body_operations.cpp
peci1: This doesn't change any semantics. The `Shape*` constructor calls `setDimensions()` internally. E.g. here: https://github.com/ros-planning/geometric_shapes/blob/c17a1de10c4befd9e220b9c79afa5df0cda85154/include/geometric_shapes/bodies.h#L454 .

Modifying this function is not needed. It doesn't bring any speedup. It just shortens the code.

src/body_operations.cpp
peci1: As above, no semantics is changed (the original `createBodyFromShape()` would call `setDimensions()` internally). It just allows utilizing the speedup using `*Dirty` functions.

rhaschke (approved):
Thanks for the clarification.

v4hn:
Thanks @peci1 . Merging.

PR #138, opened 2020-04-28
Added body operations constructShapeFromBody() and constructMarkerFromBody().

peci1:
This helps very much when you want to visualize bodies in RViz and want to be sure the bug is in your code and not in the way you create the marker.

This can be backported to melodic. The ABI-breaking functions getScaledDimensions() could be at least added as non-virtual to keep ABI.

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -224,6 +224,9 @@ public:
   /** \brief Get the dimensions associated to this body (as read from corresponding shape) */
   virtual std::vector<double> getDimensions() const = 0;
 
+  /** \brief Get the dimensions associated to this body (scaled and padded) */
+  virtual std::vector<double> getScaledDimensions() const = 0;
+
   /** \brief Set the dimensions of the body (from corresponding shape) */
   inline void setDimensions(const shapes::Shape* shape)
   {
@@ -331,6 +334,7 @@ public:
 
   /** \brief Get the radius of the sphere */
   std::vector<double> getDimensions() const override;
+  std::vector<double> getScaledDimensions() const override;
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
@@ -382,6 +386,7 @@ public:
 
   /** \brief Get the radius & length of the cylinder */
   std::vector<double> getDimensions() const override;
+  std::vector<double> getScaledDimensions() const override;
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
@@ -443,6 +448,7 @@ public:
 
   /** \brief Get the length & width & height (x, y, z) of the box */
   std::vector<double> getDimensions() const override;
+  std::vector<double> getScaledDimensions() const override;
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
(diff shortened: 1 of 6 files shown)


v4hn:
@peci1 could you please rebase this to get conflicts resolved.

peci1:
Rebased.

rhaschke (approved):

rhaschke:
@v4hn, do you want to give this a second review?

v4hn (approved):
One more cleanup from my side.
If CI succeeds, I approve.

peci1:
Should I prepare the backporting PR for melodic with non-virtual functions?

v4hn:
If you are interested in the methods in melodic, then go ahead, yes.
I guess you propose that people can handle the down-cast from `bodies::Body` on their own before using the methods?

peci1:
That's one way. Another way would be to also add the non-virtual methods to `bodies::Body` and switch using dynamic_casts or typeid. It'd be ugly, but probably not that much as there are only a few bodies defined...

v4hn:
Sounds reasonable. At least it would provide the same API.
The (minor) downside is that the new interface would require RTTI.
But I don't know whether anyone is actually interested in disabling it nowadays.

peci1:
Hmm, this could probably be handled by some CMake magic to detect RTTI and then making two different implementations (the non-RTTI one probably throwing an exception?). But I'm not sure if it's worth the hassle...

v4hn:
> But I'm not sure if it's worth the hassle...

I agree and would just ignore it. Your proposal sounds good to include in melodic.

rhaschke:
Is it really worth the effort to backport this? Which feature in the Melodic branches rely on it?

peci1:
It'd at least allow me to drop my ugly custom workarounds in http://github.com/peci1/robot_body_filter/blob/7b69ec6d5a9231adc7eb0b52c9c270903522a21c/src/utils/bodies.cpp .

rhaschke:
As long as you are the only one requiring this feature, you can simply build from master source, can't you?
UPDATE: However, if you _want_ to put this extra effort, you are, of course, welcome to do so.

v4hn:
> As long as you are the only one requiring this feature, you can simply build from master source, can't you?

I don't see why you would try to convince contributors not to provide tested backports.
With MTC you are always the one who requires compatibility with the releases and downstream compatibility is always so much easier if upstream provides the same APIs across versions...

rhaschke:
We always need to balance the extra effort for backporting (and here code different from the master branch is required to maintain ABI stability) and the benefit. I don't yet see other users than Martin utilizing the new feature in Melodic. I'm not trying to convince him to not provide the backport as I hopefully clarified in my update...

peci1:
It's not just me who's using it. The package is released into melodic and has at least three users (or people who tried it and reported issues). Thus building against master is not an option.

The extra effort is worth it for me, so I'll prepare the PR - it will just take some time because of vacations...

peci1:
And I think the logic is twisted:

> I don't yet see other users than Martin utilizing the new feature in Melodic

How could people use the new features when it has only been merged into Noetic? And there's more - I think the non-presence of `getScaledDimensions()` might have had discouraged some people from using geometric_shapes, because without this function you basically create bodies whose dimensions are unobservable from outside the library. And that makes utilizing these bodies pretty hard in downstream code (and that's the reason why I needed the ugly `#define private public` hack).

rhaschke:
People can use this feature even if they are not using a (not yet existing) binary Noetic release: They just need to build this package from source within their ROS workspace.

v4hn:
@peci1 As this just came up again in https://github.com/PR2/robot_self_filter/issues/23#issuecomment-674032618 ,
feel free to provide a melodic backport.

peci1:
It took me time, but the backport is here: #209.

PR #142, opened 2020-05-04
Added a function to create an empty body from shape type.

v4hn:
Backport of (#137).

This allows more efficient body construction when scale, padding or pose should also be set during the construction.

### DIFF
--- a/include/geometric_shapes/body_operations.h
+++ b/include/geometric_shapes/body_operations.h
@@ -46,6 +46,9 @@
 namespace bodies
 {
 /** \brief Create a body from a given shape */
+Body* createEmptyBodyFromShapeType(const shapes::ShapeType& shapeType);
+
+/** \brief Create a body from a given shape */
 Body* createBodyFromShape(const shapes::Shape* shape);
 
 /** \brief Create a body from a given shape */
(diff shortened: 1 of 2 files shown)


rhaschke:
Argh! Do you really want to apply the backporting paradigm here as well?
Applying the same set of commits on several branches is highly confusing. I quickly lose track of what was applied where.
For rviz, I'm following a merge paradigm: Everything that should become available in the current release branch, should be applied to that branch and _merged_ into the master / noetic branch.
This way, you can more easily see what was applied where:

image

So #137 should have been applied to melodic-devel and then merged into noetic-devel ...

v4hn:
I'm also fine with not backporting at all and have MoveIt `master` rely on the noetic branch of this repository.
That implies changes in `moveit` related to changes here will never be backported to moveit's `melodic-devel` branch though.

You pointed out your preferred merge strategy a number of times by now.
It relies on rather strict control and agreement among all maintainers and leads to many changes in a single merge commit if patches have to be adjusted to other changes between the branches.
For my last backport pr in `moveit` that would have lead to ~100 lines of unaccounted changes as part of a merge commit. Using cherry-picks they are now part of the individual commits which makes much more sense to me.

For `geometric_shapes` I don't really care. But I think it's a bad idea to start treating individual moveit-related repositories differently, as this just leads to confusion among maintainers and even more documentation on standard maintainance routines...

rhaschke:
Obviously both approaches have their pros and cons. I don't care a lot for this repo as well.
So, I'm going to proceed as you suggested...

PR #143, opened 2020-05-21
get rid of dynamic_cast

v4hn:
This library does not need RTTI as all bodies & shapes are marked with `type`.

The EXPECT_TRUE in the test degenerates to true for all minimally-optimizing compilers.
Replaced it with something meaningful.

should be backported to melodic.

### DIFF
--- a/src/body_operations.cpp
+++ b/src/body_operations.cpp
@@ -160,9 +160,11 @@ void bodies::computeBoundingSphere(const std::vector<const bodies::Body*>& bodie
   unsigned int vertex_count = 0;
   for (auto body : bodies)
   {
-    const bodies::ConvexMesh* conv = dynamic_cast<const bodies::ConvexMesh*>(body);
-    if (!conv)
+    if (!body || body->getType() != shapes::MESH)
       continue;
+    // MESH type implies bodies::ConvexMesh
+    const bodies::ConvexMesh* conv = static_cast<const bodies::ConvexMesh*>(body);
+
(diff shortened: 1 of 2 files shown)


src/body_operations.cpp
```diff
@@ -160,9 +160,11 @@ void bodies::computeBoundingSphere(const std::vector<const bodies::Body*>& bodie
   unsigned int vertex_count = 0;
   for (auto body : bodies)
   {
-    const bodies::ConvexMesh* conv = dynamic_cast<const bodies::ConvexMesh*>(body);
-    if (!conv)
+    if (body->type != MESH)
```
peci1: First test if `body != nullptr`. The old check would rule such cases out, whether the suggested change would segfault.

v4hn:
> First test if `body != nullptr`.
> The old check would rule such cases out, whether the suggested change would segfault.

True, I updated the patch.
I'm not perfectly happy with silently skipping nullptr's in the list,
but the patch was not meant to change behavior.

rhaschke:
@v4hn, have you seen that your code is broken? For example, you need to use `type_` instead of `type`.

v4hn:
> @v4hn, have you seen that your code is broken? For example, you need to use `type_` instead of `type`.

Actually, CI was partially just broken for this test for network issues. But yes, there was still one more namespace mistake I just fixed.

rhaschke (approved):

PR #144, opened 2020-05-22
Factored out the checkIsometry() helper function into a standalone header library.

peci1:
As suggested in https://github.com/ros-planning/moveit/pull/1964 .

I made the checkIsometry() function available also in release mode, so that the check can be used at other places than inside the assert macro.

### DIFF
--- a/include/geometric_shapes/check_isometry.h
+++ b/include/geometric_shapes/check_isometry.h
@@ -1,0 +1,116 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Martin Pecka
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the Willow Garage nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Martin Pecka, Robert Haschke */
+
+#ifndef GEOMETRIC_SHAPES_CHECK_ISOMETRY_H
+#define GEOMETRIC_SHAPES_CHECK_ISOMETRY_H
+
+#include <algorithm>
+#include <assert.h>
+#include <iostream>
+#include <math.h>
+#include <Eigen/Core>
+#include <Eigen/Geometry>
+
+/** \brief This file provides functions and macros that can be used to verify that an Eigen::Isometry3d is really an
+ * isometry. Eigen itself doesn't do the checks because they're expensive. If the isometry object is constructed in
+ * a wrong way (e.g. from an AngleAxisd with non-unit axis), it can represent a non-isometry. But some methods in the
+ * Isometry3d class perform isometry-specific operations which return wrong results when called on a non-isometry. E.g.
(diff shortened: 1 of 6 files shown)


include/geometric_shapes/check_isometry.h
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// To be able to use the more efficient Eigen::Transform::linear() instead of rotation(),
+// we need to check the user has really passed an isometry. To avoid runtime costs,
+// this check is only done as assert, which gets compiled-out in release builds
+#ifdef NDEBUG
+/** \brief Assert that the given transform is an isometry (no-op in release mode). */
+// this is functionally a no-op, but prevents unused variable warnings
+// clang-format off
+#define ASSERT_ISOMETRY(transform) do { (void)sizeof(transform);} while (0);
```
rhaschke: Why so complicated? Wouldn't `(void)sizeof(transform)` suffice to silence unused parameter warnings?

include/geometric_shapes/check_isometry.h
```diff
@@ -0,0 +1,143 @@
...
+ * \param floatPrecision The float precision to which float numbers are considered.
+ * \return The desired number of decimal digits (after decimal dot). The result will always be at least 6.
+ * \note Used as a parameter for std::setprecision().
+ */
+inline int getPrintPrecision(const double floatPrecision)
+{
+  return std::max(6, static_cast<int>(std::abs(std::log10(floatPrecision))) + 1);
+}
+}  // namespace
+
```
rhaschke: Do we really need this? Why can't we simply go with default (scientific) formatting as before?

include/geometric_shapes/check_isometry.h
```diff
@@ -0,0 +1,143 @@
...
+ * \brief Call a failed assertion if the given transform is not an isometry.
+ * \param transform The possibly non-isometric transform.
+ * \param precision Precision to which the transform has to correspond to an isometry (element-wise).
+ */
+inline void assertIsometry(const Eigen::Isometry3d& transform, const double precision = CHECK_ISOMETRY_PRECISION)
+{
+  if (!::bodies::checkIsometry(transform, precision))
+    assert(!"Invalid isometry transform");
+}
+
```
rhaschke: I suggest avoiding a function here, because file information will not resolve to the correct line number in that case.
I have to admit that the previous solution didn't correctly resolve as well (due to my suggestion).

rhaschke (commented):
From my point of view, the checkIsometry function is somewhat over-engineered, allowing to pass the precision as well.

peci1:
The rationale for passing the precision and setting it one order "higher" than the precision used for comparison was because I didn't like errors like:

> The given transform is not an isometry! It's linear part involves scaling: 1 1 1

But if it seems over-engineered to you, I can make the precision hard-coded which would simplify the code.

include/geometric_shapes/check_isometry.h
peci1: So you suggest the `checkIsometry` should be a macro instead?

include/geometric_shapes/check_isometry.h
peci1: I'll try it. This is what I found on stackoverflow, though.

include/geometric_shapes/check_isometry.h
rhaschke: No, checkIsometry should stay a function as originally suggested by me.
But, the content of `assertIsometry` should be moved into the macro `ASSERT_ISOMETRY`.
Essentially, it is important to have all `assert` statements within a macro to correctly resolve the line number.

rhaschke:
> But if it seems over-engineered to you, I can make the precision hard-coded which would simplify the code.

I was expecting the `1 1 1` error output as the underlying reason for this change. What about a message like this:
`"The linear part involves a scaling, which deviates from identity by %f", max(abs(scaling - ones())`

peci1:
I'm not for introducing other possible rounding/representation errors. I was also thinking about subtracting ones, but in the end I think that showing the raw computed values is more valuable.

include/geometric_shapes/check_isometry.h
peci1: Fixed.

include/geometric_shapes/check_isometry.h
peci1: Good, your suggestion works.

rhaschke (approved):
I still don't like the complexity of the code, but I approve. It's up to @v4hn to ask for more simplification if necessary.

peci1:
Thanks. I'm open to discussion.

rhaschke:
> I'm not for introducing other possible rounding/representation errors.

I don't think you introduce new rounding errors by the proposed approach. You will just focus on the _deviations_ from the expected values.

peci1:
I was more concerned about representation errors of IEEE floats. In 32 bits, There are approx 2^30 numbers between 0 and 1, and only 2^23 between 1 and 2. But I don't think that if you take a number from 1-2 range and subtract one, you'd have a guarantee that you'll find a corresponding normal float in the 0-1 range. 

However, since these values are anyways computed by SVD, they probably accumulate so many errors during the computation that caring about the representation of the result can be seen as a negligible issue...

So, okay, you convinced me to subtract ones and use the scientific format :)

include/geometric_shapes/check_isometry.h
peci1: Solved, displaying difference from the expected values in scientific format.

include/geometric_shapes/check_isometry.h
```diff
@@ -0,0 +1,118 @@
...
+ * isometry. Eigen itself doesn't do the checks because they're expensive. If the isometry object is constructed in
+ * a wrong way (e.g. from an AngleAxisd with non-unit axis), it can represent a non-isometry. But some methods in the
+ * Isometry3d class perform isometry-specific operations which return wrong results when called on a non-isometry. E.g.
+ * for isometries, .linear() and .rotation() should be the same, but for non-isometries, the result of .linear()
+ * contains also the scaling factor, whether .rotation() is only the rotation part.
+ *
+ * These checks are primarily meant to be performed only in debug mode (via the ASSERT_ISOMETRY macro), but you can call
+ * checkIsometry() even in release mode. This check should be mainly performed on transforms input by the user.
+ */
+namespace bodies
```
v4hn: I don't see why this generic function should live in `bodies::`.
Any complains about moving this out of the namespace?

peci1: I think having it in no namespace would be dangerous... But a different namespace might work... I though about misusing `Eigen`, but I'm not sure it's a good idea.

v4hn: Given the generic nature of this function, I don't see a problem with the global namespace. Hijacking `Eigen` is wrong. Going for a `utils` namespace is not much better than the global namespace, `geometric_shapes::checkIsometry` seems rather long (but still better than the current version).
Either way, the main interface of this code, `ASSERT_ISOMETRY`, of course does not use a namespace..

peci1: > Either way, the main interface of this code, `ASSERT_ISOMETRY`, of course does not use a namespace..

That's a fair argument, we're polluting the global namespace anyways...

peci1: I removed the namespace.

v4hn (approved):

PR #145, opened 2020-05-27
Allow a larger tolerance on matrix scaling

AndyZe:
Related to https://github.com/ros-planning/moveit/issues/2112

### DIFF
--- a/include/geometric_shapes/check_isometry.h
+++ b/include/geometric_shapes/check_isometry.h
@@ -57,7 +57,7 @@
 
 #ifndef CHECK_ISOMETRY_PRECISION
 /** \brief The default precision to which the transform has to correspond to an isometry. */
-#define CHECK_ISOMETRY_PRECISION 1e-6
+#define CHECK_ISOMETRY_PRECISION Eigen::NumTraits<float>::dummy_precision()
 #endif
 
 /**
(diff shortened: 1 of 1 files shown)


peci1:
The question is, whether you are ready to bear the consequences of non-isometries. There are now places in moveit code which depend on the inputs to be exact isometries. If you pass general affine transforms (or worse), the results will be wrong and you won't be notified of that.

In my point of view, it's better to catch these villains earlier and notify the user that he should do something about the computations to retain the isometric properties.

But I agree that the precision introduced in that commit was an arbitrary choice not backed by any real geometrical interpretation or limitation.

v4hn:
I would not label the default precision for Eigen fuzzy comparisons of double values "an arbitrary choice" as we use doubles for our isometries. Reducing the precision to factor 10 of the noise observed in an arbitrary code base on the other hand *is* pretty arbitrary...

If we change this, I would prefer to go for `Eigen::NumTraits<float>::dummy_precision()` instead, assuming we want users to handle their isometries in floats in their pipelines.

I am not convinced that this is a good idea in general though, because we rely on Eigen which is perfectly justified to expect Isometries up to its own precision requirement.

@AndyZe before the commit you referenced, the normalization was done *implicitly* at a very high cost (SVD of the linear part). If you just reduce the factor here, it will not be done *at all* anymore.

AndyZe:
MoveIt is probably multiplying transforms repeatedly, right? To calculate link transforms? Maybe that is what is causing this error.

@v4hn, good call about using `Eigen::NumTraits<float>::dummy_precision()`. I'll try that

AndyZe:
Yes, the default float precision works for me.

rhaschke (commented):
@AndyZe: Is this still an issue for you? If not, please close.
As we use double precision for all transforms, we should use double precision for checks as well.

PR #147, opened 2020-06-05
Merge 'noetic' into 'dashing-devel'

henningkayser:
Note: I would like to move to eloquent if this succeeds (already enabled in CI) and we can probably drop dashing support. Should we just switch to a `ros2` branch for simplicity? I just don't know if this is feasible with upcoming foxy support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,10 @@ Note: Bodies for meshes compute the convex hull of those meshes in order to prov
 
 Note: [shape_tools](https://github.com/ros-planning/shape_tools) package was recently merged into this package
 
+Note: `bodies::Box::corner1_` was renamed to `minCorner_` and `bodies::Box::corner2_` to `maxCorner_` in Noetic.
+
+Note: `bodies::ConvexMesh::MeshData` was made implementation-private in Noetic and is no longer accessible from the `.h` file.
+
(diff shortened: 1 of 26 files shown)


README.md
```diff
@@ -15,6 +15,10 @@ Note: Bodies for meshes compute the convex hull of those meshes in order to prov
 
 Note: [shape_tools](https://github.com/ros-planning/shape_tools) package was recently merged into this package
 
+Note: `bodies::Box::corner1_` was renamed to `minCorner_` and `bodies::Box::corner2_` to `maxCorner_` in Noetic.
```
peci1: Not sure if the mention of Noetic would be useful for ROS 2 people...

README.md
```diff
@@ -15,6 +15,10 @@ Note: Bodies for meshes compute the convex hull of those meshes in order to prov
...
+
+Note: `bodies::ConvexMesh::MeshData` was made implementation-private in Noetic and is no longer accessible from the `.h` file.
```
peci1: dtto

peci1 (commented):
Looks good from my side.

README.md
henningkayser: of course, thanks

PR #149, opened 2020-06-07
Port to qhull_r

jspricke:


jspricke:
Unit tests started to segfault when integrating 0.7.0 into Debian. As the non reentrant qhull version is deprecated anyhow, I ported it to qhull_r.

src/bodies.cpp
```diff
@@ -933,20 +913,24 @@ void bodies::ConvexMesh::useDimensions(const shapes::Shape* shape)
   static FILE* null = fopen("/dev/null", "w");
 
   char flags[] = "qhull Tv Qt";
-  int exitcode = qh_new_qhull(3, mesh->vertex_count, points, true, flags, null, null);
+  qhT qh_qh;
+  qhT *qh= &qh_qh;
```
v4hn:
```suggestion
  qhT* qh = &qh_qh;
```

v4hn (commented):
This change apparently breaks support with ROS kinetic. I guess `libqhull_r` was not available in Ubuntu 16.04?

Sounds like it's time to move the minimum supported distribution for the MoveIt stack to melodic.

Please have a look at CI, the melodic builds fail because of missing function symbols:
https://travis-ci.com/github/ros-planning/geometric_shapes/jobs/345628627#L615

jspricke:
>This change apparently breaks support with ROS kinetic. I guess `libqhull_r` was not available in Ubuntu 16.04?

Yeah, looks like I only added it in 2017:

http://changelogs.ubuntu.com/changelogs/pool/universe/q/qhull/qhull_2015.2-4/changelog

>Sounds like it's time to move the minimum supported distribution for the MoveIt stack to melodic.

Already, I moved all robots to Melodic like half a year ago ;). Will you 
it?

>Please have a look at CI, the melodic builds fail because of missing function symbols:
>https://travis-ci.com/github/ros-planning/geometric_shapes/jobs/345628627#L615

Will have a look tonight.

>> @@ -933,20 +913,24 @@ void bodies::ConvexMesh::useDimensions(const shapes::Shape* shape)
>   static FILE* null = fopen("/dev/null", "w");
>
>   char flags[] = "qhull Tv Qt";
>-  int exitcode = qh_new_qhull(3, mesh->vertex_count, points, true, flags, null, null);
>+  qhT qh_qh;
>+  qhT *qh= &qh_qh;
>
>```suggestion
>  qhT* qh = &qh_qh;
>```

Oh yeah, c&p from Qhull, fixed now.

jspricke:
>Please have a look at CI, the melodic builds fail because of missing function symbols:
>https://travis-ci.com/github/ros-planning/geometric_shapes/jobs/345628627#L615

Fixed.

v4hn (approved):
I approve this change. qhull is a pain in the a** and using the reentrant interface makes sense for `geometric_shapes`.

Please note that this breaks with Ubuntu 16.04, effectively breaking support of the whole MoveIt `master` stack in ROS kinetic. Before merge, this should be discussed at the maintainer meeting.
A few other changes can also be done at the same time (c++17 dependency, `boost::python`, ?), but we still had ROS kinetic users building `master` a while back.
If we merge this, CI setups for `kinetic` have to be removed in multiple repositories.

v4hn:
@tylerjw @rhaschke I know you already released this package into noetic, but I would propose to get this patch in before the main MoveIt release and re-release geometric_shapes along with it.

Strictly speaking, this is not necessary for noetic compatibility, but rather is the first change that would be relevant "beyond noetic".
Also it is a good patch to occasion disabling kinetic builds in CI :-)

cmake/FindQhull.cmake
```diff
@@ -12,30 +12,29 @@ set(QHULL_DEBUG_NAME qhull_rd)
...
           NAMES libqhull_r.h
           HINTS "${QHULL_ROOT}" "$ENV{QHULL_ROOT}" "${QHULL_INCLUDE_DIR}"
-          PATHS "$ENV{PROGRAMFILES}/QHull" "$ENV{PROGRAMW6432}/QHull" 
+          PATHS "$ENV{PROGRAMFILES}/QHull" "$ENV{PROGRAMW6432}/QHull"
           PATH_SUFFIXES libqhull_r)
 
 set(QHULL_HEADER "${QHULL_HEADER}" CACHE INTERNAL "QHull header" FORCE )
 
 if(QHULL_HEADER)
-  get_filename_component(qhull_header ${QHULL_HEADER} NAME_WE)
```
rhaschke, line 30: As far as I can see, this `qhull_header` variable is not used.

rhaschke (commented):
I added a commit to remove an unused cmake variable (by editor also cleaned up trailing whitespace).
I updated .travis.yml to build on Noetic + Melodic only.

rhaschke:
Attempting to fix the Travis issues, I messed up this PR even more. Sorry for that.
I suggest reverting my last commit (`Updating .travis.yml` - only @jspricke can do this) and merge this as is.
To fix Travis (and clang-format-10), I have prepared new commits running on [Travis right now](https://travis-ci.com/github/rhaschke/geometric_shapes/builds/182073116).

jspricke:
@rhaschke removed your commit

rhaschke:
Thanks. Merging as is - with broken Travis, which will get fixed via another PR.

PR #154, opened 2020-07-29
Update CMakeLists.txt to include octomap headers

ruffsl:
Looks like octomap isn't a properly implemented ament package, so `ament_target_dependencies` doesn't necessary pickup on the correct location for the included header files for the library. This explicitly adds the includes directory using the env exported by the octomap's cmake config, as prescribed here:

https://wiki.ros.org/octomap#Using_OctoMap

I suspect this went under the radar for so long as CI or maintainers where building this packages using released system installed deps for octomap, rather than from a source build of octomap in the same workspace.

cc @mikaelarguedas

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,10 @@ if(HAVE_QHULL_2011)
   add_definitions(-DGEOMETRIC_SHAPES_HAVE_QHULL_2011)
 endif()
 
-include_directories(include)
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
See a suggestion for a  slightly less disruptive approach at https://github.com/ros-planning/geometric_shapes/pull/155

henningkayser:

PR #155, opened 2020-07-29
Use uppercase octomap to fix header not found errors 

mikaelarguedas:
Alternative to https://github.com/ros-planning/geometric_shapes/pull/154

The octomap exported CMake variable when find_package'd are using the uppercase name (e.g. OCTOMAP_INCLUDE_DIRS).
This means the lowercase versions (e.g. octomap_INCLUDE_DIRS) are empty and this cause headers not to be found when compiling geometric_shapes.

This PR uses the uppercase name to find octomap and passes OCTOMAP to `ament_target_dependencies`, this way amnt uses the uppercase CMake variables.

@ruffsl can you check if this fixes your issue?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ find_package(Boost REQUIRED system filesystem)
 find_package(console_bridge REQUIRED)
 find_package(eigen3_cmake_module REQUIRED)
 find_package(Eigen3 REQUIRED)
-find_package(octomap REQUIRED)
+find_package(OCTOMAP REQUIRED)
 find_package(ament_cmake REQUIRED)
 find_package(eigen_stl_containers REQUIRED)
 find_package(random_numbers REQUIRED)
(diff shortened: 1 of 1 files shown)


ruffsl:
Just tested this, as it works for me :+1: . I thought of doing this, but think I saw from git blame that this was changed to lower case previously, but the PR is a bit odd https://github.com/ros-planning/geometric_shapes/pull/124

mikaelarguedas:
Yeah it's a bit odd indeed, rosdep works off the content of package.xml not the CMakeLists :thinking:

henningkayser:
@mikaelarguedas @ruffsl Thanks for looking into this. The upper case version is already provided by https://github.com/ros-planning/geometric_shapes/pull/153 which still has to be reviewed and merged. Please let me know if you have any issues with that branch.

ruffsl:
@henningkayser , just tested your https://github.com/ros-planning/geometric_shapes/pull/153 with my https://github.com/ros-planning/moveit2/pull/233 and tests are passing.

https://app.circleci.com/pipelines/github/ruffsl/moveit2/13/workflows/d4135583-4298-4cda-94d3-5097b8b6a69c/jobs/39

henningkayser:

PR #156, opened 2020-07-30
[ros2] trim boost dependencies

mikaelarguedas:
In an [ongoing effort](https://discourse.ros.org/t/generating-dev-and-runtime-artefacts-from-ros-packages/12448) to reduce the footprint on the target systems, it is preferred to declare specific dependencies instead of generic ones.

Requires ros/rosdistro#25995


mikaelarguedas:
Updated to retrigger CI now that underlying rosdep key PR has been merged

henningkayser (approved):
Thanks!

PR #157, opened 2020-08-19
add soname version to library

tylerjw:
This is adding the soname version to the library to enable future releases of Noetic to brake the ABI in a nicer way.  This is specifically because Noetic is the last ROS1 release and we can no longer depend on future versions of ROS1 if we want to break the ABI.


v4hn (approved):
Looks good from my perspective.

As noetic matures, we might want to head for version-independent SONAMEs though (so that we only break linkage if there is actually ABI breakage.

PR #159, opened 2020-09-25
Merge branch 'noetic-devel' into ros2

henningkayser:

PR #162, opened 2020-09-25
Merge updates from noetic-devel into ros2

rhaschke:
Replacement for #159.


henningkayser (approved):
I think this requires https://github.com/ros-planning/moveit_ci/pull/118 to succeed CI

henningkayser:
@rhaschke this still requires some ament_lint fixes

rhaschke:
I'm reluctant to hardcode the package version. Is there no way to inherit this from package.xml?
https://github.com/ros-planning/geometric_shapes/blob/a637c76e3fc78e146994ddd68a20440aecd94fee/CMakeLists.txt#L2

rhaschke:
@henningkayser, gcc 9.3 seems to complain about redundant qhull declarations. Hence we need to disable warnings.
I filed an issue upstream as well: https://github.com/qhull/qhull/issues/78

henningkayser:
> I'm reluctant to hardcode the package version. Is there no way to inherit this from package.xml?
> https://github.com/ros-planning/geometric_shapes/blob/a637c76e3fc78e146994ddd68a20440aecd94fee/CMakeLists.txt#L2

I agree, we could set the version [this way](https://github.com/rhaschke/geometric_shapes/pull/1)

> @henningkayser, gcc 9.3 seems to complain about redundant qhull declarations. Hence we need to disable warnings.
> I filed an issue upstream as well: [qhull/qhull#78](https://github.com/qhull/qhull/issues/78)

That's odd. I didn't run into this, but this should definitely be fixed.

JafarAbdi (approved):

PR #164, opened 2020-11-25
[fix] cmake install command

tylerjw:
This makes the install command conform to the standard here: https://index.ros.org/doc/ros2/Tutorials/Ament-CMake-Documentation/

Specifically what it fixes though is the EXPORT in the install allows this repo to be used in an installed underlay.


rhaschke (approved):
Thanks for providing the ament doc link. No more questions :wink:

henningkayser (approved):

PR #169, opened 2021-03-18
Migrate to GitHub actions for ROS2

tylerjw:
These changes are the same we are now using on warehouse_ros.  I'll test this in my own fork and link it.


rhaschke (commented):
What about the other linters:
- `catkin_lint` or the pendant in ROS2
- `clang-tidy`

Are they part of `industrial_ci`?

Please also include the `CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unused-parameter"`. Maybe, we should move those to the top-level cmake file, so that issues don't come at a surprise on CI only?
(I always prefer to keep control over such flags as a user, though, i.e. not enforcing them via cmake.)

tylerjw:
Here is it passing: https://github.com/tylerjw/geometric_shapes/pull/1/checks

tylerjw:
>     * `catkin_lint` or the pendant in ROS2

An equivalent doesn't exist for ROS2.  There is ament_lint_cmake, but it does such a tiny amount of linting compared to what the really nice tool catkin_lint did.

>     * clang-tidy

I'll work on adding this.  This is a special animal for ros2 in that we'll need to either depend on the latest source in ament_lint to make it usable for clang-tidy (and even then I think I still need to fix some things there and then we'd depend on my PR).  There is a industrial_ci feature for running clang-tidy, I'll see if I can get that to work instead.


> Please also include the `CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unused-parameter"`. Maybe, we should move those to the top-level cmake file, so that issues don't come at a surprise on CI only?
> (I always prefer to keep control over such flags as a user, though, i.e. not enforcing them via cmake.)

I'll figure this out.  I'm also conflicted about this.  On one hand it is really annoying when you have to do extra iterations because the CI system is testing with a different build configuration.  At the same time I really don't like cmake files that change the global build flags (as then they affect things that depend on this package).  Would adding these to each target in the cmake be a reasonable compromise.

rhaschke:
> Would adding these to each target in the cmake be a reasonable compromise.

Yes, I think that's the way to go. You can make those flags private as well, so that they don't clutter downstream builds.

tylerjw:
I think I figured out how to specify the ccache stuff and a bit more about the github cache action.  If you get an exact hit on the key for the cache it will not push back to the cache.  As a result I added the git hash to the key so it wouldn't push back if you are just re-running the same thing, but otherwise I think it should always push back into the cache if there was a change.

The other thing was how I was using the home enviroment variable.  It was creating a relative path to a read-only part of the file system.  The ~ seems to work correctly (because the parser of the github yaml doesn't expand it) and expands into the home directory in the docker image started by industrial_ci and in the host system (even though they have different home paths).

tylerjw:
This is ready for review, the job completed here and is now passing (with clang-tidy, ament_lint_cmake, and ccache): https://github.com/tylerjw/geometric_shapes/pull/1/checks?check_run_id=2141721421

Edit: It is re-running because I changed the last commit and force pushed to remove the branch name from the key as it seems redundant now that I'm using the commit hash in the key.  I believe it will pass though with no issues.

Update: It passed again.  See link.

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,36 @@
...
+name: BuildAndTest
+
+# This determines when this workflow is run
+on:
+  # When ros2 branch is pushed to
+  push:
+    branches: [ ros2 ]
+  # When there is a pull request against ros2
+  pull_request:
+    branches: [ ros2 ]
```
rhaschke: I guess this can be simplified: On other branches, the workflow isn't defined at all (or differently).
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-a-list-of-events
```suggestion
# Triggers the workflow on push or pull request events
on: [push, pull_request]
```

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  industrial_ci:
+    strategy:
+      matrix:
+        env:
+          - {ROS_DISTRO: foxy, ROS_REPO: main}
+          - {ROS_DISTRO: foxy, ROS_REPO: testing}
+          - {ROS_DISTRO: foxy, ROS_REPO: main, CLANG_TIDY: pedantic}
```
rhaschke: What about combining these?
```suggestion
          - {ROS_DISTRO: foxy, ROS_REPO: testing, CLANG_TIDY: pedantic}
```

geometric_shapes.repos
```diff
@@ -1,8 +1,4 @@
 repositories:
-  geometric_shapes:
-    type: git
-    url: https://github.com/ros-planning/geometric_shapes
-    version: ros2
   random_numbers:
```
rhaschke, line 2: `random_numbers` still isn't released?

tylerjw:
I'm sorry to keep iterating on this.  Looking at how clang-tidy is run by industrial_ci the job with it and without it are redundant.  The clang-tidy test does all the normal build and test steps too.  I'll drop the non-clang-tidy job and that should be my last change here.

tylerjw:
OK, now this is ready for review.  I further rebated to remove what were essentially fixup commits so the history has a clearer set of changes to look over.

.github/workflows/industrial_ci_action.yml
```diff
@@ -0,0 +1,32 @@
...
+    env:
+      UPSTREAM_WORKSPACE: geometric_shapes.repos
+      CCACHE_DIR: ~/.ccache
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ccache-${{ matrix.env.ROS_DISTRO }}-${{ matrix.env.ROS_REPO }}-${{ github.sha }}
```
rhaschke, line 28: I see the following warning for the 3rd job:
```
Post job cleanup.
Unable to reserve cache with key ccache-foxy-main-190fd59788645cbe4195b0056d862e1f21363f2d, another job may be creating this cache.
```
I think this is due to the fact that the 1st and 3rd build jobs try to create the same cache key: The key doesn't distinguish between CLANG_TIDY on/off. Should be resolved with my suggestion above. In general, you need to take care considering all build-relevant env variables.

rhaschke (commented):
You rebased in the middle of my review...
LGTM.

geometric_shapes.repos
tylerjw: no, sadly, it is on the things we have planned to do over the next couple weeks though.

tylerjw:
> You rebased in the middle of my review...
> LGTM.

I'm sorry, but it looks like I read your mind on some of these.  Thank you again for reviewing these.  I'm really excited to finally be done with Travis if only on these smaller repos.

rhaschke:
I think you merged this prematurely.

tylerjw:
> I think you merged this prematurely.

I'm sorry... I thought your `LGTM` was an edit to say you approved.

tylerjw:
I see now I didn't respond to all your feedback.  I missed the simplification of the `on` line.  :facepalm:

rhaschke:
> I'm really excited to finally be done with Travis if only on these smaller repos.

Thanks for your effort on this. Great job! I need to turn tp other things now (you were flooding me with review requests :wink:)
I guess we agreed on a common config. Hence, I think you can proceed on your own (w/o detailed review) now. Just as a reminder, this is the set of constraints I have particularly checked:
- Limit to max. 2 commits (switch to gh actions + apply formatting changes)
- Check that gh actually succeeds w/o warnings
- Check that all previous Travis checks are included again.

tylerjw:
Thank you for helping me review these.  I'll be careful about those things and will get others to review as I try to get this done for our few other small repos.

PR #177, opened 2021-04-07
Update CI Workflows [melodic-devel]

tylerjw:
one more, copy pasta of the others


.github/workflows/industrial_ci_action.yaml
```diff
@@ -0,0 +1,70 @@
...
+# For troubleshooting, see readme (https://github.com/ros-industrial/industrial_ci/blob/master/README.rst)
+
+name: BuildAndTest
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - noetic-devel
```
rhaschke: This only applies to the melodic-devel branch
```suggestion
```

rhaschke (approved):
See minor fixup. Please merge yourself if this is addressed and CI is happy.

.github/workflows/industrial_ci_action.yaml
tylerjw: done, will do

PR #182, opened 2021-04-27
Fix export depends

tylerjw:
This fixes the build issue with moveit2 on main.  I believe this is the root cause of the issue.

* export all dependencies that have working cmake using variable for dependency list and export depends
* sort the find_package section that can be sorted (eigen is a special child and the order matters for it)


CMakeLists.txt
```diff
@@ -100,11 +110,6 @@ if(BUILD_TESTING)
   ament_lint_auto_find_test_dependencies()
 endif()
 
-install(
-  DIRECTORY include/
-  DESTINATION include
-)
-
```
tylerjw, line 107: these are a duplicate of line 121

v4hn (approved):
clear improvements over ths previously quite broken state of that file...

Issue #185, opened 2021-04-30
--- stderr: geometric_shapes/src/bodies.cpp:44: warning: redundant redeclaration(s)


newcanopies:
Hi, should this error be logged when building `geometric_shapes `  and does it affect package functionality?

```
--- stderr: geometric_shapes
In file included from /home/user/ws_ROS2/src/moveit2/geometric_shapes/src/bodies.cpp:44:
/usr/include/libqhull_r/libqhull_r.h:1057:9: warning: redundant redeclaration of ‘void qh_printsummary(qhT*, FILE*)’ in same scope [-Wredundant-decls]
 1057 | void    qh_printsummary(qhT *qh, FILE *fp);
      |         ^~~~~~~~~~~~~~~
/usr/include/libqhull_r/libqhull_r.h:1024:9: note: previous declaration of ‘void qh_printsummary(qhT*, FILE*)’
 1024 | void    qh_printsummary(qhT *qh, FILE *fp);
      |         ^~~~~~~~~~~~~~~
/usr/include/libqhull_r/libqhull_r.h:1099:6: warning: redundant redeclaration of ‘void qh_meminit(qhT*, FILE*)’ in same scope [-Wredundant-decls]
 1099 | void qh_meminit(qhT *qh, FILE *ferr);
      |      ^~~~~~~~~~
In file included from /usr/include/libqhull_r/libqhull_r.h:31,
                 from /home/user/ws_ROS2/src/moveit2/geometric_shapes/src/bodies.cpp:44:
/usr/include/libqhull_r/mem_r.h:218:6: note: previous declaration of ‘void qh_meminit(qhT*, FILE*)’
  218 | void qh_meminit(qhT *qh, FILE *ferr);
      |      ^~~~~~~~~~
In file included from /home/user/ws_ROS2/src/moveit2/geometric_shapes/src/bodies.cpp:44:
/usr/include/libqhull_r/libqhull_r.h:1100:6: warning: redundant redeclaration of ‘void qh_memfreeshort(qhT*, int*, int*)’ in same scope [-Wredundant-decls]
 1100 | void qh_memfreeshort(qhT *qh, int *curlong, int *totlong);
      |      ^~~~~~~~~~~~~~~
In file included from /usr/include/libqhull_r/libqhull_r.h:31,
                 from /home/user/ws_ROS2/src/moveit2/geometric_shapes/src/bodies.cpp:44:
[...]

PR #186, opened 2021-04-30
Add qhull as SYSTEM dependency to suppress compiler warnings

JafarAbdi:
Fix: https://github.com/ros-planning/geometric_shapes/issues/185

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,8 +97,8 @@ ament_target_dependencies(${PROJECT_NAME}
 
   # We don't export these dependencies because their cmake is broken
   ASSIMP
-  QHULL
 )
+ament_target_dependencies(${PROJECT_NAME} SYSTEM QHULL)
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,13 +92,13 @@ add_library(${PROJECT_NAME} SHARED
 )
 set_target_properties(${PROJECT_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
 target_compile_options(${PROJECT_NAME} PRIVATE ${PROJECT_COMPILE_OPTIONS})
-ament_target_dependencies(${PROJECT_NAME}
+ament_target_dependencies(${PROJECT_NAME} SYSTEM
   ${THIS_PACKAGE_EXPORT_DEPENDS}
 
   # We don't export these dependencies because their cmake is broken
   ASSIMP
+  QHULL
 )
-ament_target_dependencies(${PROJECT_NAME} SYSTEM QHULL)
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -97,8 +97,8 @@ ament_target_dependencies(${PROJECT_NAME}
 
   # We don't export these dependencies because their cmake is broken
   ASSIMP
-  QHULL
 )
+ament_target_dependencies(${PROJECT_NAME} SYSTEM QHULL)
```
rhaschke: Shouldn't all these dependencies be `SYSTEM`?

JafarAbdi: This should also work, I updated the PR to make all dependencies `SYSTEM`

v4hn: We had problems with overlayed workspaces in ROS1/catkin multiple times because `SYSTEM` was used for packages that can be built in the same workspace and should overlay system versions. This can be problematic because `-isystem` will not just silence warnings, but also change the order of these includes to be behind all `-I` includes.

I'm not sure about the situation in ament/ros2 though.

JafarAbdi: > I'm not sure about the situation in ament/ros2 though.

I'm pretty sure it similar to ROS1/catkin since they will have the same include flags, but I never had this problem before

v4hn: It's somewhat involved, but the problem comes together because:
- ROS include paths for install workspaces are the same for packages in the same workspace.
- `-isystem` includes always override `-I` includes if both are specified for the same folder

So it can happen that two dependent packages both include the path from their shared workspace, but one of them declares it via `-isystem`. In that case the path will only be considered after **all** `-I` includes. When working with overlayed workspaces, it's quite likely though that one of them contains a different version of the other package and you will get broken builds.

That's why I would always only declare dependencies that are not involved in ROS (and that will never be built in a ROS workspace) as `SYSTEM`

tylerjw: Would it then make sense to split up the dependencies that are ROS from those that are and only declare the non-ROS ones as SYSTEM?

rhaschke: That a very valid argument from Michael. Given that tricky situation, I agree with Tyler to only declare non-ROS packages as SYSTEM. Sorry for the noise.

CMakeLists.txt
```diff
@@ -92,7 +92,7 @@ add_library(${PROJECT_NAME} SHARED
 )
 set_target_properties(${PROJECT_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
 target_compile_options(${PROJECT_NAME} PRIVATE ${PROJECT_COMPILE_OPTIONS})
-ament_target_dependencies(${PROJECT_NAME}
+ament_target_dependencies(${PROJECT_NAME} SYSTEM
   ${THIS_PACKAGE_EXPORT_DEPENDS}
```
tylerjw:
```suggestion
ament_target_dependencies(${PROJECT_NAME}
  ${THIS_PACKAGE_EXPORT_DEPENDS_ROS}
ament_target_dependencies(${PROJECT_NAME} SYSTEM
  ${THIS_PACKAGE_EXPORT_DEPENDS_EXTERNAL}
```

Maybe you can split them up this way with two different lists.  Make sure you update the ament_export_dependencies line to include both.

v4hn (approved):

rhaschke (approved):

PR #189, opened 2021-05-30
Fix building on Windows with MSVC

Ace314159:
The ROS2 branch fails to compile on Windows with MSVC. 

There were 3 things I had to change.
1. The warning flags (`-Wextra`, `-Wwrite-strings`, etc.) aren't available, so I just got rid of those when using MSVC.
2. `M_PI` is not defined unless `_USE_MATH_DEFINES` is also defined.
3. `__cplusplus` is not set to the correct value unless the `/Zc:__cplusplus` compiler flag is used. However, simply adding this compiler flag still doesn't fix the issue for any other packages that include the files. However, `_MSVC_LANG` seems to be set the correct value, so I used that macro to perform the check.

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -37,10 +37,6 @@
 #ifndef GEOMETRIC_SHAPES_BODIES_
 #define GEOMETRIC_SHAPES_BODIES_
 
-#if __cplusplus <= 199711L && _MSVC_LANG <= 199711L
-#error This header requires at least C++11
-#endif
-
 #define _USE_MATH_DEFINES
 #include "geometric_shapes/aabb.h"
 #include "geometric_shapes/shapes.h"
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -37,7 +37,7 @@
 #ifndef GEOMETRIC_SHAPES_BODIES_
 #define GEOMETRIC_SHAPES_BODIES_
 
-#if __cplusplus <= 199711L
+#if __cplusplus <= 199711L && _MSVC_LANG <= 199711L
 #error This header requires at least C++11
 #endif
 
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -41,6 +41,7 @@
 #error This header requires at least C++11
 #endif
 
+#define _USE_MATH_DEFINES
 #include "geometric_shapes/aabb.h"
 #include "geometric_shapes/shapes.h"
 #include <eigen_stl_containers/eigen_stl_containers.h>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,8 +17,6 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     -Wredundant-decls
     -Wno-unused-parameter
   )
-else()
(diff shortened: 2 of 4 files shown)


nkalupahana:
@henningkayser or anyone else, can we get this merged in? Again, this fixes my workflow, and it's important for all of the ROS2 windows users.

include/geometric_shapes/bodies.h
```diff
@@ -37,7 +37,7 @@
 #ifndef GEOMETRIC_SHAPES_BODIES_
 #define GEOMETRIC_SHAPES_BODIES_
 
-#if __cplusplus <= 199711L
+#if __cplusplus <= 199711L && _MSVC_LANG <= 199711L
```
henningkayser: Shouldn't this be `||`?

henningkayser:
@Ace314159 could you update your branch?

include/geometric_shapes/bodies.h
Ace314159: It should be `&&` because we don't want to enter the if. On windows, the first condition is always true due to the incorrect value of `__cplusplus`, so with `||` it would always the enter the if and give an error.

I tested on WSL with gcc and this still compiles even though _MSVC_LANG isn't defined.

Ace314159:
Thanks for looking at this! I have updated my branch.

include/geometric_shapes/bodies.h
henningkayser: Makes sense... On second thought, these checks and errors seem redundant to me. I don't think it's realistic that someone uses C++98 with ROS2. @tylerjw what do you think?

tylerjw: I don't think that's reasonable considering that ROS1 itself standardizes on C++11 and ROS2 on C++14.

henningkayser: So let's remove these checks I'd say

tylerjw: Are we maybe not telling the compiler we want C++11 or newer somehow and MSVCC defaults to C++98?

Ace314159: Even though the compiler is for C++11 and newer, the macro value is incorrect for MSVC without the `Zc:__cplusplus` compiler flag. This [Microsoft blog post](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/) explains this in more detail.

tylerjw: I read the blog post and I'm still confused.  Is this just a way to detect having an old msvcc compiler based on them taking longer to update this macro?

henningkayser (approved):

include/geometric_shapes/bodies.h
Ace314159: Even on newer MSVC compilers the macro is still wrong. The only way to tell the compiler to use the correct value is through the use of the compiler flag. I didn't just add that compiler flag to the compile options because that would require all packages that include those header files to also add that compiler option. Instead, MSVC uses a different macro (`_MSVC_LANG`) that has the same value as `__cplusplus` on other compilers. And this doesn't need any compiler flag to be accurate.

However, based on the discussion above, I thought I should remove those checks. If not, I can revert that commit.

tylerjw: Ok, I think I understand now.  Thank you.  I think removing those checks is just fine as ROS doesn't support old C++ standards so we shouldn't need to check for them anyway.

tylerjw (approved):

PR #191, opened 2021-06-10
Fix memory leak in createMeshFromBinary

wxmerkt:
When using `createMeshFromBinary` `scene` gets created as a raw pointer but never deleted even after the pointer goes out of scope. This resulted in approximately 1.1 GB of memory leaked when loading some of our URDFs. Explicitly deleting the object after use fixes this problem.

### DIFF
--- a/src/mesh_operations.cpp
+++ b/src/mesh_operations.cpp
@@ -261,7 +261,10 @@ Mesh* createMeshFromBinary(const char* buffer, std::size_t size, const Eigen::Ve
   // so they must be delayed until after clearing the root node transform above.
   importer.ApplyPostProcessing(aiProcess_OptimizeMeshes | aiProcess_OptimizeGraph);
 
-  return createMeshFromAsset(scene, scale, hint);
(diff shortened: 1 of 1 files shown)


tylerjw:
You closed this, did you find this fix wasn't needed?

v4hn:
The in-source documentation for `Assimp::Importer::ReadFileFromMemory` in Assimp 5.0.1 (also earlier versions) states:

     * The returned data is intended to be
     * read-only, the importer object keeps ownership of the data and will
     * destroy it upon destruction. [...]
     * The previous scene will be deleted during this call.

So this patch is invalid. If there is a memory leak here, it might be in Assimp itself.

wxmerkt:
I closed the PR as the patch turned out to be invalid - there is a leak, however, this fix wouldn't be correct as `AssimpImporter` takes care of cleaning up its own memory. If I find a fix, I'll make a new PR.

v4hn:
> there is a leak [...] If I find a fix, I'll make a new PR.

Happy debugging! We're of course looking forward to the patch if the leak is in our code base.

PR #195, opened 2021-06-16
[ROS2] Add prerelease tests

vatanaksoytezer:


vatanaksoytezer:
@tylerjw @henningkayser Galactic will fail on this one until we get a random_numbers release on galactic

.github/workflows/prerelease.yaml
```diff
@@ -0,0 +1,27 @@
+# This config uses industrial_ci (https://github.com/ros-industrial/industrial_ci.git).
+# For troubleshooting, see readme (https://github.com/ros-industrial/industrial_ci/blob/master/README.rst)
+
+name: pre-release
+
+on:
+  workflow_dispatch:
+  pull_request:
```
henningkayser: Could you remove pull_request for now?

.github/workflows/prerelease.yaml
```diff
@@ -0,0 +1,27 @@
...
+      fail-fast: false
+      matrix:
+        distro: [foxy, galactic, rolling]
+
+    env:
+      ROS_DISTRO: ${{ matrix.distro }}
+      PRERELEASE: true
+      BASEDIR: ${{ github.workspace }}/.work
+
+    name: "${{ matrix.distro }}"
```
henningkayser, line 21: Prefix with "Prerelease Test" or so?

vatanaksoytezer: Sure.

.github/workflows/prerelease.yaml
vatanaksoytezer: I left this since it is failing. Would you like to merge this anyway? If so I can remove

.github/workflows/prerelease.yaml
vatanaksoytezer: For the convention I can write this to all repos or left it like that. GHA Already kinda adds this though: 
image

henningkayser: If gha already adds something, then it's fine like this. I just noticed that the checks for this PR only list the distro names.

.github/workflows/prerelease.yaml
henningkayser: Yeah, let's merge this anyway

vatanaksoytezer: Removed it.

henningkayser (approved):

PR #200, opened 2021-07-17
 Changed retriever.h to retriever.hpp in mesh_operations.cpp

Drojas251:
Error found when building ROS2 Rolling and Moveit2 from source. the resource_retriever ros2 branch removed depreciated retriever.h header file

### DIFF
--- a/src/mesh_operations.cpp
+++ b/src/mesh_operations.cpp
@@ -44,7 +44,7 @@
 #include <float.h>
 
 #include <console_bridge/console.h>
-#include <resource_retriever/retriever.h>
+#include <resource_retriever/retriever.hpp>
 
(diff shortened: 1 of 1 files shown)


vatanaksoytezer (approved):

vatanaksoytezer:
@Drojas251 thanks for reporting, indeed they seem to remove this header with https://github.com/ros/resource_retriever/pull/63. This change will not compile on Foxy, so I think we would have to branch off for foxy at this point. Pinging @tylerjw and @henningkayser to let the CI run and branch off for foxy.

vatanaksoytezer:
@tylerjw can you approve the CI for this PR? We should merge this if CI passes.

tylerjw (approved):

Issue #166, opened 2021-03-03
foxy build error linking boost libbrary


zuxinrui:
**Platform:** ubuntu 20.04 foxy
When build the package from source using ros2 branch, something wrong with linking the boost library:

--- stderr: geometric_shapes
/usr/bin/ld: CMakeFiles/test_loaded_meshes.dir/test_loaded_meshes.cpp.o: in function `CompareMeshVsPrimitive::SetUp()':
test_loaded_meshes.cpp:(.text._ZN22CompareMeshVsPrimitive5SetUpEv[_ZN22CompareMeshVsPrimitive5SetUpEv]+0x13f): undefined reference to `boost::filesystem::path::operator/=(boost::filesystem::path const&)'
collect2: error: ld returned 1 exit status
make[2]: *** [test/CMakeFiles/test_loaded_meshes.dir/build.make:158: test/test_loaded_meshes] Error 1
make[1]: *** [CMakeFiles/Makefile2:422: test/CMakeFiles/test_loaded_meshes.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
/usr/bin/ld: CMakeFiles/test_point_inclusion.dir/test_point_inclusion.cpp.o: in function `MeshPointContainment_Pr2Forearm_Test::TestBody()':
test_point_inclusion.cpp:(.text+0x710b): undefined reference to `boost::filesystem::path::operator/=(boost::filesystem::path const&)'
/usr/bin/ld: CMakeFiles/test_point_inclusion.dir/test_point_inclusion.cpp.o: in function `MeshPointContainment_Basic_Test::TestBody()':
test_point_inclusion.cpp:(.text+0x17216): undefined reference to `boost::filesystem::path::operator/=(boost::filesystem::path const&)'
collect2: error: ld returned 1 exit status
make[2]: *** [test/CMakeFiles/test_point_inclusion.dir/build.make:158: test/test_point_inclusion] Error 1
make[1]: *** [CMakeFiles/Makefile2:277: test/CMakeFiles/test_point_inclusion.dir/all] Error 2
/usr/bin/ld: CMakeFiles/test_body_operations.dir/test_body_operations.cpp.o: in function `Bodies_ConstructShapeFromBodyMesh_Test::TestBody()':
test_body_operations.cpp:(.text+0x2346): undefined reference to `boost::filesystem::path::operator/=(boost::filesystem::path const&)'
[...]

PR #205, opened 2021-10-28
Fix cmake such that Boost::filesystem is exported properly

jlack1987:
This fixes https://github.com/ros-planning/geometric_shapes/issues/166. Ultimately [this issue](https://github.com/ament/ament_cmake/issues/282) result in the `Boost` components not getting exported with an `ament_target_dependencies` and `ament_export_dependencies` calls.

Before this fix attempting to link against this pkg resulted in,
```
Target "mylib" links to target "Boost::filesystem"
  but the target was not found.  Perhaps a find_package() call is missing for
  an IMPORTED target, or an ALIAS target is missing?
```

After this fix it builds properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,4 +126,4 @@ install(DIRECTORY include/ DESTINATION include)
 ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
 ament_export_dependencies(${THIS_PACKAGE_EXPORT_DEPENDS})
 
-ament_package()
+ament_package(CONFIG_EXTRAS "cmake/ConfigExtras.cmake")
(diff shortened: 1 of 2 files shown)


henningkayser (commented):
Thank you for fixing this. Small nit: could you rename the cmake file to "ConfigExtras.cmake" and also include it to replace `find_package(Boost.. )`? Mostly for consistency with the other packages and to keep Boost dependencies in a single spot.

jlack1987:
@henningkayser I went ahead and fixed the file name, but I don't think that file should be included. Is there somewhere else you have seen that does this?

jlack1987:
Is there any chance this could get cherry picked into `foxy`/`galactic` since it's a bug that makes it tough to use this package?

JafarAbdi:
For some reason, GitHub decided to merge this PR by me only pushing to your branch :/

JafarAbdi:
@jlack1987 Do you mind testing the latest `ros2` branch .?

JafarAbdi:
> Is there any chance this could get cherry picked into foxy/galactic since it's a bug that makes it tough to use this package?

Yep, I'll do right now

PR #206, opened 2021-11-01
Fix cmake such that Boost::filesystem is exported properly

JafarAbdi:
Reopen https://github.com/ros-planning/geometric_shapes/pull/205, see my comment https://github.com/ros-planning/geometric_shapes/pull/205#issuecomment-956245216 :(


JafarAbdi:
@jlack1987 please review https://github.com/jlack1987/geometric_shapes/pull/1

jlack1987:
@JafarAbdi your change there doesn't fix the issue. The problem is that packages depending on `geometric_shapes` can't properly link against it because the boost components aren't properly exported

JafarAbdi:
> @JafarAbdi your change there doesn't fix the issue. The problem is that packages depending on `geometric_shapes` can't properly link against it because the boost components aren't properly exported

I don't think this should be the case, we have been using this method with moveit2 and linking against it without any problem, do you mind trying a clean build .?

jlack1987:
I'm betting you're picking up `Boost::filesystem` from some other package you depend on then and that's how it's working. If you go into one of your packages that depends on `geometric_shapes` and do `message(STATUS ${geometric_shapes_LIBRARIES})` you'll see that it does not include `Boost::filesystem`. That's how I discovered that `geometric_shapes` was where the problem was.

JafarAbdi:
Out of curiosity, how are you linking your code against `geometric_shapes`, is it possible to share your CMakeLists .? thanks!

JafarAbdi:
> I'm betting you're picking up `Boost::filesystem` from some other package you depend on then and that's how it's working. If you go into one of your packages that depends on `geometric_shapes` and do `message(STATUS ${geometric_shapes_LIBRARIES})` you'll see that it does not include `Boost::filesystem`. That's how I discovered that `geometric_shapes` was where the problem was.

Actually, I do see it in `${geometric_shapes_LIBRARIES}` tested on rolling

jlack1987:
I can't really share the entire `CMakeLists` but it's nothing special, just
```
add_library(${PROJECT_NAME}_collision SHARED src/collision_monitor.cpp)
ament_target_dependencies(${PROJECT_NAME}_collision geometric_shapes)
```
I have more libraries in the `ament_target_dependencies` line in practice, but the way I found out that `geometric_shapes` was the problem was I commented out all the code from the library and started removing dependencies from that line and it failed to build until I removed `geometric_shapes` so then I went digging and have addressed [this issue](https://github.com/ament/ament_cmake/issues/199) with other libraries so I knew what to look for

henningkayser:
From previous PR discussion: I think we should include the cmake config instead of the [Boost include](https://github.com/ros-planning/geometric_shapes/pull/206/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR51) like [in moveit_core](https://github.com/ros-planning/moveit2/blob/main/moveit_core/CMakeLists.txt#L13). That way, all Boost components are defined in one place which is easier to maintain and less error-prone.

jlack1987:
Yeah i'll leave that up to you all. Just pointing out that the way it exists right now unless that `Boost::filesystem` lib is picked up by some other dependency you can't depend on this package.

jlack1987:
Made recommended changes. Let me know if there's anything else I need to do. Thanks

tylerjw (approved):
I've had to make this change in other places before.  Thank you for doing this.

jlack1987:
Yeah np, any chance we could cherry-pick this back into `foxy`?

tylerjw:
I don't see why not.  Please just open this change to the foxy branch and I'll happily merge it.

PR #209, opened 2021-11-22
[melodic backport] Added body operations constructShapeFromBody() and constructMarkerFromBody().

peci1:
This is a backport of #138 to melodic. All tests pass on my side in both debug and release modes. I used the fake non-virtual polymorphy for `Body::getScaledDimensions()` to retain feature parity with noetic and keep melodic ABI unchanged.

I added a special test case for calling `constructShapeFromBody()` on a `bodies::Body` pointer rather than on one of the downstream types. I can port this test to the noetic branch, too, to test that the polymorphic behavior is doing what it is expected to do. I've actually noticed that neither dynamic, nor static casts are needed in the places edited by commit bc1cf5ea43f6b8d09cfd5d53f68f418cdae9defd - as `getScaledDimensions()` is polymorphic (noetic) or fake polymorphic (melodic), calling it directly on a `Body` instance should be good enough. I also did this simplification in this PR, and can transfer it to noetic-devel, too (but it's just a cosmetic issue).

Accepting this backport would allow me to drop a `#define private public` hack from https://github.com/peci1/robot_body_filter/blob/master/src/utils/bodies.cpp (and would also allow me to drop `constructShapeFromBody()` and `constructMarkerFromBody()` implementations in robot_body_filter and use the geometric_shapes versions).

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -199,6 +199,9 @@ public:
   /** \brief Get the dimensions associated to this body (as read from corresponding shape) */
   virtual std::vector<double> getDimensions() const = 0;
 
+  /** \brief Get the dimensions associated to this body (scaled and padded) */
+  virtual std::vector<double> getScaledDimensions() const = 0;
+
   /** \brief Set the dimensions of the body (from corresponding shape) */
   void setDimensions(const shapes::Shape* shape);
 
@@ -313,6 +316,7 @@ public:
 
   /** \brief Get the radius of the sphere */
   std::vector<double> getDimensions() const override;
+  std::vector<double> getScaledDimensions() const override;
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
@@ -372,6 +376,7 @@ public:
 
   /** \brief Get the radius & length of the cylinder */
   std::vector<double> getDimensions() const override;
+  std::vector<double> getScaledDimensions() const override;
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
@@ -444,6 +449,7 @@ public:
 
   /** \brief Get the length & width & height (x, y, z) of the box */
   std::vector<double> getDimensions() const override;
+  std::vector<double> getScaledDimensions() const override;
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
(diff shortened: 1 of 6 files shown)


peci1:
> @peci1 As this just came up again in https://github.com/PR2/robot_self_filter/issues/23#issuecomment-674032618 ,
feel free to provide a melodic backport.

_Originally posted by @v4hn in https://github.com/ros-planning/geometric_shapes/issues/138#issuecomment-674088013_

@v4hn could you please have a look at this until Melodic times out? :)

v4hn (approved):
... That PR ended up in the backlog (as many others...)

I believe we [bump the library SONAME](https://github.com/ros-planning/geometric_shapes/blob/noetic-devel/CMakeLists.txt#L102) in this repository, so ABI breakage should not be a concern and I would be fine with simplifying your logic again to a virtual interface.

That aside I approve.
@rhaschke, you were the second review for the original PR, so maybe you can spare a moment?

peci1:
Thanks for the review! I would definitely love to drop the ABI workaround.

I was just checking #138 and found out that Robert has added commit https://github.com/ros-planning/geometric_shapes/pull/138/commits/68e67408fc827ec04f7fefb49c84fee246b801c1 which is not in this PR. I think it should be included.

I've also inspected your additional commit https://github.com/ros-planning/geometric_shapes/pull/138/commits/bc1cf5ea43f6b8d09cfd5d53f68f418cdae9defd which has turned `dynamic_casts` into `static_casts` in `constructShapeFromBody()`. However, I'm now not sure why a cast is needed at all in this case. Function `getScaledDimensions()` is virtual in #138, so it should not be needed to downcast to particular body types to call it. But I don't remember well if there wasn't some other reason. I guess it would not even save a vtable lookup as even the downcasted functions are virtual.

rhaschke:
> Function `getScaledDimensions()` is virtual in #138, so it should not be needed to downcast to particular body types to call it. 

I agree.

> I guess it would not even save a vtable lookup as even the downcasted functions are virtual.

Not yet. But https://github.com/ros-planning/geometric_shapes/pull/225 should avoid the vtable lookup.

rhaschke:
I agree with Michael, that we could drop the ABI compatibility.

peci1:
I've removed the ABI compatibility commit and added the two additional commits from #138. I'd wait for #225 to be merged, cherry-pick the commit here, and then this PR could be ready. What do you think?

rhaschke (approved):

peci1:
Thanks!

PR #210, opened 2021-11-22
[noetic] Added bodies::Body::computeBoundingBox (oriented box version)

peci1:
This is a noetic version of #105. I basically just cherry-picked the commits from #105 and resolved merge conflicts.

I also figured out this PR breaks ABI by adding a virtual function, so I again made it fake virtual. It's a pity it will be such in both noetic and melodic, but there's not much more that can be done.

I also made use of the fact that FCL 0.6 is available in noetic and relayed AABB computation in `bodies::AABB::extendWithTransformedBox()` directly to FCL. This change is a bit off-topic, but does not hurt and offers to get potential future improvements from FCL. I made this change in a separate commit, so if it is not desired, I can get rid of it.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,4 +1,4 @@
-<package>
+<package format="3">
   <name>geometric_shapes</name>
   <version>0.7.3</version>
   <description>Generic definitions of geometric shapes and bodies.</description>
@@ -19,7 +19,8 @@
   <build_depend>eigen</build_depend>
   <build_depend>eigen_stl_containers</build_depend>
   <build_depend>libconsole-bridge-dev</build_depend>
-  <build_depend>libfcl-dev</build_depend>
+  <build_depend condition="$ROS_DISTRO != noetic">libfcl-dev</build_depend>
+  <build_depend condition="$ROS_DISTRO == noetic">fcl</build_depend>
   <build_depend>libqhull</build_depend>
   <build_depend>octomap</build_depend>
   <build_depend>pkg-config</build_depend>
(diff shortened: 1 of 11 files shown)


peci1:
The FCL search logic in CMake might look complicated, but that's what I had to do in https://github.com/peci1/robot_body_filter/commit/db08750759d5dd93822451a7c2981368c6bd187b#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a in order to get it correctly working on Melodic and Noetic on both Ubuntu and Debian.

rhaschke (approved):
Looks good to me. I'm fine to include FCL in general. I would even drop ABI compatibility and risk some release friction in favour of cleaner code.

peci1:
> I would even drop ABI compatibility

Should I remove the ABI compatibility commit, then?

rhaschke:
> Should I remove the ABI compatibility commit, then?

Yes, please.

rhaschke:
@ros-planning/moveit-maintainers, this is awaiting the 2nd approval for more than a month...

package.xml
```diff
@@ -19,6 +19,8 @@
   <build_depend>eigen</build_depend>
   <build_depend>eigen_stl_containers</build_depend>
   <build_depend>libconsole-bridge-dev</build_depend>
+  <build_depend condition="$ROS_DISTRO != noetic">libfcl-dev</build_depend>
```
simonschmeisser, line 22: does this include apply for ROS2 as well? or should it be `==melodic` else ...

src/aabb.cpp
```diff
@@ -36,14 +36,19 @@
 
 #include <geometric_shapes/aabb.h>
 
+#include <fcl/config.h>
+#if FCL_MINOR_VERSION > 5
```
simonschmeisser: I doubt it but there might be a `1.0.0` release at some point

simonschmeisser:
Do we actually intend to backport this to `melodic-devel`? If not it could be simplified a lot by dropping support for fcl 0.5 ... or maybe not as the libfcl package in Ubuntu Focal is still 0.5.0 while the ros-noetic-fcl is at 0.6.1

package.xml
peci1: no idea how ROS 2 handles this... I haven't dived into it yet... so maybe just melodic/noetic would be better

peci1:
I saw that versions of FCL can range from 0.5 to 0.6 even on newer systems depending on many things, so I'd rather keep it compatible with both. And yes, I'd like to backport this to melodic.

package.xml
simonschmeisser: Debian bullseye contains 0.6.1, debian sid 0.7.1(and therefore also in Ubuntu 22.04) and there was no ros based releases except for noetic so this sounds like the correct approach

simonschmeisser (approved):

src/aabb.cpp
peci1: I added the major version checks, too.

simonschmeisser:
Looks good now!

Do you want to squash the second commit into the first before merging?

peci1:
> Do you want to squash the second commit into the first before merging?

Do you mean to not have a commit where build is broken?

simonschmeisser:
> > Do you want to squash the second commit into the first before merging?
> 
> Do you mean to not have a commit where build is broken?

Yes and because the commit is somewhat "boring" :wink: and looks like a fixup

peci1:
Okay, squashed ;)

peci1:
Great, thanks! Now there's #105 which could be used for the melodic backport. Should I help with it somehow?

PR #213, opened 2021-12-22
2.1.1 (Foxy)

tylerjw:
This is a problem because 2.1.1 was released for Rolling and Galactic but never for foxy.  I'm not sure what to do about that as we should probably release these changes into foxy.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,18 @@
 Changelog for package geometric_shapes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2.1.1 (2021-12-22)
+------------------
+* Fix cmake such that Boost::filesystem is exported properly (`#206 <https://github.com/ros-planning/geometric_shapes/issues/206>`_) (`#211 <https://github.com/ros-planning/geometric_shapes/issues/211>`_)
+  Co-authored-by: Jafar Abdi <cafer.abdi@gmail.com>
+  Co-authored-by: Jordan Lack <jlack@houstonmechatronics.com>
+* Enable CI in foxy (`#208 <https://github.com/ros-planning/geometric_shapes/issues/208>`_)
+* CI: Update pat-s/always-upload-cache
+* CI for Foxy only
+* Add prerelease tests (`#195 <https://github.com/ros-planning/geometric_shapes/issues/195>`_)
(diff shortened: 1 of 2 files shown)


rhaschke:
Why do you need a separate foxy branch at all? As far as I can see (but I checked a few files only), the code differences are limited to the copyright notice. What about simply merging and continuing with a singly ros2 branch?

tylerjw:
Based on the email I got from the buildfarm it is yet another one of the .h/.hpp incldes.  I'm going to make a PR against the ros2 branch that should make this build for foxy.

CHANGELOG.rst
```diff
@@ -2,6 +2,18 @@
 Changelog for package geometric_shapes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2.1.1 (2021-12-22)
+------------------
+* Fix cmake such that Boost::filesystem is exported properly (`#206 <https://github.com/ros-planning/geometric_shapes/issues/206>`_) (`#211 <https://github.com/ros-planning/geometric_shapes/issues/211>`_)
+  Co-authored-by: Jafar Abdi <cafer.abdi@gmail.com>
```
henningkayser, line 8: Co-authors can be dropped here and listed under "Contributors", imo. Also, do we want to keep/drop CI changes from the changelog in the future?

henningkayser (approved):
Looks good, fixing my comment is optional. Feel free to merge and push the tag.

henningkayser:
> Based on the email I got from the buildfarm it is yet another one of the .h/.hpp incldes. I'm going to make a PR against the ros2 branch that should make this build for foxy.

Maybe we were a little to fast with creating these branches. In the future they might still be useful, and syncing the changes here should be trivial.

tylerjw:
@henningkayser look at the other PR for adding support for Foxy to the ros2 branch.  I think we should close this as it has a SemVar issue and we should be able to release from R/G/F all from the ros2 branch after that other PR.  Then after the release is synced we can delete this branch (buildfarm doesn't keep old versions anyway).

henningkayser:
> @henningkayser look at the other PR for adding support for Foxy to the ros2 branch. I think we should close this as it has a SemVar issue and we should be able to release from R/G/F all from the ros2 branch after that other PR. Then after the release is synced we can delete this branch (buildfarm doesn't keep old versions anyway).

Yeah, just wanted to say that I noticed the other one right now ;)

PR #214, opened 2021-12-23
Foxy support via __has_include

tylerjw:
`__has_include` is a C++17 feature [(Reference)](https://en.cppreference.com/w/cpp/preprocessor/include).  We can support foxy from the ros2 branch with this change.  The downside to this is foxy doesn't officially support C++17.  This is the reason I originally created the foxy branch.  If we are ok using C++17 on Foxy we can support all three current ros2 versions (Foxy, Galactic, Rolling) from one branch.  This will also fix the semvar issues created by lot leaving room for more foxy versions (as we can just release foxy from the ros2 branch then).

### DIFF
--- a/src/mesh_operations.cpp
+++ b/src/mesh_operations.cpp
@@ -38,7 +38,13 @@
 #include <float.h>
 
 #include <console_bridge/console.h>
+
+// Shim so this works on foxy/galactic/rolling
+#if __has_include(<resource_retriever/retriever.hpp>)
 #include <resource_retriever/retriever.hpp>
+#else
+#include <resource_retriever/retriever.h>
+#endif
 
 #include <assimp/scene.h>
 #include <assimp/Importer.hpp>
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -24,6 +24,8 @@ jobs:
             ROS_REPO: main
           - ROS_DISTRO: galactic
             ROS_REPO: testing
+          - ROS_DISTRO: foxy
+            ROS_REPO: testing
     env:
       UPSTREAM_WORKSPACE: geometric_shapes.repos
       AFTER_SETUP_UPSTREAM_WORKSPACE: 'vcs pull $BASEDIR/upstream_ws/src'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,9 @@
 cmake_minimum_required(VERSION 3.5)
 project(geometric_shapes)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 
 # Set compile options


src/mesh_operations.cpp
```diff
@@ -38,7 +38,13 @@
 #include <float.h>
 
 #include <console_bridge/console.h>
+
+// Shim so this works on foxy/galactic/rolling
+#if __has_include(<tf2_eigen/tf2_eigen.hpp>)
```
rhaschke: Why don't you test for `resource_retriever/retriever.hpp` directly?

rhaschke (commented):
Alternatively, to avoid C++17, you could test for the existence of one or the other header file in cmake and add a corresponding define.

src/mesh_operations.cpp
tylerjw: Because I copied this from somewhere in moveit without reading it :facepalm: 

Fixed that.

tylerjw:
> Alternatively, to avoid C++17, you could test for the existence of one or the other header file in cmake and add a corresponding define.

Do you have an example for how to do that?

tylerjw:
I'm biased in that I think it is really silly that Foxy is using C++14 when the two Tier 1 OS's have C++17 by default in their compilers (Ubuntu 20.04 and Windows).  In my opinion, the version of C++ that ROS uses should just follow the defaults in the compiler for whatever the first-class targets are.

tylerjw:
The natural objection is something along the lines of "but some embedded target might not have a new enough compiler".  I would challenge those who are saying that to go find those compilers and make a list of them somewhere publically.  In my experience, every embedded target I've used had compilers that support newer versions of C++ about as fast as the ones on linux.  I challenge anyone to find a microcontroller (in the currently supported version of the microcontroller from the manufacturer) that you cannot find a compiler with C++20 support.  It is hard enough to get people to upgrade ROS who are on linux... I can't imagine there are enough people using an EOL micro-controller wanting to upgrade their ROS version but not their micro-controller and compiler.

tylerjw:
Sorry for the off-topic rant.  I'm ok with a define from cmake if you have an example of how to search for a header in cmake and set the variable.

v4hn:
[Example](https://github.com/ros-planning/moveit_task_constructor/blob/062c127b86316e02f7a19de3d90ccfbd285f6902/core/CMakeLists.txt#L33-L40). That was the (rather verbose) way to go for many years. :-)

I believe @rhaschke just meant to point out an alternative. Personally I don't see a problem with using `__has_include`. It *is* a very neat feature. If, against all odds, someone should complain about it, it's straight-forward to patch it out.

tylerjw:
> [Example](https://github.com/ros-planning/moveit_task_constructor/blob/062c127b86316e02f7a19de3d90ccfbd285f6902/core/CMakeLists.txt#L33-L40). That was the (rather verbose) way to go for many years. :-)
> 
> I believe @rhaschke just meant to point out an alternative. Personally I don't see a problem with using `__has_include`. It _is_ a very neat feature. If, against all odds, someone should complain about it, it's straight-forward to patch it out.

I like the solution here better.  If either of you cares about this strongly I'm happy to change it though.  I am just happy we can fix the SemVar problem I created and get rid of that foxy branch.

v4hn (approved):

PR #223, opened 2022-05-31
[ros2] WIP: Remove random_numbers dependency

sjahr:
### Description
This PR aims to remove the random number dependency from geometric_shapes. To do so, the cpp std library `<random>` is used directly. Instead of the random_number::RandomNumberGenerator, a simpler one is defined in the header `random_number_utils.hpp`. 

Furthermore, the API of `samplePointInside` is changed by removing the random number generator argument. I did this to aviod the need for the user to create an extra RandomNumberGenerator instance  when using this function

Generating random quaternions can be done with the constructor from Eigen itself. For example like this:
```
Eigen::Quaterniond rand_quaternion = Eigen::Quaterniond::UnitRandom();
```

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -225,10 +225,7 @@ public:
      Sometimes multiple attempts need to be generated.
      The function terminates with failure (returns false) after \e max_attempts attempts.
      If the call is successful (returns true) the point is written to \e result */
-  virtual bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const;
-
-  /** \brief Alternative version of samplePointInside function that allows using a customly seeded random number generator */
-  virtual bool samplePointInside(std::seed_seq const& seed_sequence, unsigned int max_attempts,
+  virtual bool samplePointInside(std::function<double(double, double)> random_value, unsigned int max_attempts,
                                  Eigen::Vector3d& result) const;
 
   /** \brief Compute the bounding radius for the body, in its current
@@ -305,7 +302,8 @@ public:
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
-  bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const override;
+  bool samplePointInside(std::function<double(double, double)> random_value, unsigned int max_attempts,
+                         Eigen::Vector3d& result) const override;
   void computeBoundingSphere(BoundingSphere& sphere) const override;
   void computeBoundingCylinder(BoundingCylinder& cylinder) const override;
   void computeBoundingBox(AABB& bbox) const override;
@@ -356,7 +354,8 @@ public:
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
-  bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const override;
+  bool samplePointInside(std::function<double(double, double)> random_value, unsigned int max_attempts,
+                         Eigen::Vector3d& result) const override;
   void computeBoundingSphere(BoundingSphere& sphere) const override;
   void computeBoundingCylinder(BoundingCylinder& cylinder) const override;
   void computeBoundingBox(AABB& bbox) const override;
@@ -417,7 +416,8 @@ public:
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
-  bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const override;
+  bool samplePointInside(std::function<double(double, double)> random_value, unsigned int max_attempts,
+                         Eigen::Vector3d& result) const override;
   void computeBoundingSphere(BoundingSphere& sphere) const override;
   void computeBoundingCylinder(BoundingCylinder& cylinder) const override;
   void computeBoundingBox(AABB& bbox) const override;
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -227,6 +227,10 @@ public:
      If the call is successful (returns true) the point is written to \e result */
   virtual bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const;
 
+  /** \brief Alternative version of samplePointInside function that allows using a customly seeded random number generator */
+  virtual bool samplePointInside(std::seed_seq const& seed_sequence, unsigned int max_attempts,
+                                 Eigen::Vector3d& result) const;
+
   /** \brief Compute the bounding radius for the body, in its current
       pose. Scaling and padding are accounted for. */
   virtual void computeBoundingSphere(BoundingSphere& sphere) const = 0;
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -34,8 +34,8 @@
 #define _USE_MATH_DEFINES
 #include "geometric_shapes/aabb.h"
 #include "geometric_shapes/shapes.h"
+#include "geometric_shapes/random_number_utils.hpp"
 #include <eigen_stl_containers/eigen_stl_containers.h>
-#include <random_numbers/random_numbers.h>
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 #include <memory>
@@ -225,8 +225,7 @@ public:
      Sometimes multiple attempts need to be generated.
      The function terminates with failure (returns false) after \e max_attempts attempts.
      If the call is successful (returns true) the point is written to \e result */
-  virtual bool samplePointInside(random_numbers::RandomNumberGenerator& rng, unsigned int max_attempts,
-                                 Eigen::Vector3d& result) const;
+  virtual bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const;
 
   /** \brief Compute the bounding radius for the body, in its current
       pose. Scaling and padding are accounted for. */
@@ -302,8 +301,7 @@ public:
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
-  bool samplePointInside(random_numbers::RandomNumberGenerator& rng, unsigned int max_attempts,
-                         Eigen::Vector3d& result) const override;
+  bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const override;
   void computeBoundingSphere(BoundingSphere& sphere) const override;
   void computeBoundingCylinder(BoundingCylinder& cylinder) const override;
   void computeBoundingBox(AABB& bbox) const override;
@@ -354,8 +352,7 @@ public:
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
-  bool samplePointInside(random_numbers::RandomNumberGenerator& rng, unsigned int max_attempts,
-                         Eigen::Vector3d& result) const override;
+  bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const override;
   void computeBoundingSphere(BoundingSphere& sphere) const override;
   void computeBoundingCylinder(BoundingCylinder& cylinder) const override;
   void computeBoundingBox(AABB& bbox) const override;
@@ -416,8 +413,7 @@ public:
 
   bool containsPoint(const Eigen::Vector3d& p, bool verbose = false) const override;
   double computeVolume() const override;
-  bool samplePointInside(random_numbers::RandomNumberGenerator& rng, unsigned int max_attempts,
-                         Eigen::Vector3d& result) const override;
+  bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const override;
   void computeBoundingSphere(BoundingSphere& sphere) const override;
   void computeBoundingCylinder(BoundingCylinder& cylinder) const override;
   void computeBoundingBox(AABB& bbox) const override;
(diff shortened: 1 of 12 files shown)


include/geometric_shapes/random_number_utils.hpp
```diff
@@ -0,0 +1,85 @@
...
+      std::random_device random_device;
+      std::generate_n(std::data(seed_data), std::size(seed_data), std::ref(random_device));
+      std::seed_seq sequence(std::begin(seed_data), std::end(seed_data));
+      std::mt19937 generator(sequence);
+      return generator;
+    }() } {};
+  ~RandomNumberGenerator() = default;
+
+public:
+  static RandomNumberGenerator& getInstance()
```
v4hn: Please leave a way to seed the generator.

include/geometric_shapes/bodies.h
```diff
@@ -225,8 +225,7 @@ class Body
      Sometimes multiple attempts need to be generated.
      The function terminates with failure (returns false) after \e max_attempts attempts.
      If the call is successful (returns true) the point is written to \e result */
-  virtual bool samplePointInside(random_numbers::RandomNumberGenerator& rng, unsigned int max_attempts,
-                                 Eigen::Vector3d& result) const;
+  virtual bool samplePointInside(unsigned int max_attempts, Eigen::Vector3d& result) const;
```
v4hn: I'm happy to see these new methods for simplified use, but please keep the option to specify a seeded RNG.
MoveIt is hard enough to use deterministically already.

test/test_bounding_box.cpp
```diff
@@ -80,18 +80,9 @@ TEST(SphereBoundingBox, Sphere2)
...
-  Eigen::Quaterniond quat;
-
   for (size_t i = 0; i < 10; ++i)
   {
-    gen.quaternion(quatData);
-    quat.x() = quatData[0];
-    quat.y() = quatData[1];
-    quat.z() = quatData[2];
-    quat.w() = quatData[3];
+    auto quat = Eigen::Quaterniond::UnitRandom();
```
v4hn: From what I see this is sadly not seedable (at least not without defining your own numeric trait in Eigen).
So I would prefer to keep a version somewhere that is seedable.

include/geometric_shapes/random_number_utils.hpp
sjahr: I added an additional constructor and getInstance function with the option to pass a seed as argument

test/test_bounding_box.cpp
sjahr: In this case, UnitRandom() should do the same as the unseeded  random_numbers::RandomNumberGenerator. From a viewpoint of feature completeness, a RandomNumberGenerator implementation should provide this function but my goal was only to replace the functionality that is used. I don't think such a function in the  header I used  wont be used if not here

include/geometric_shapes/bodies.h
sjahr: @v4hn Thanks for this feedback, I changed the API again to give the user the option to pass custom random values to these functions by passing a lambda that generates random values. This way the interface is flexible and we don't need to pass a library-specific RandomNumberGenerator that might not cover the functionality the user wishes. I'd like to hear your opinion on this

test/test_bounding_box.cpp
henningkayser: Eigen can (and must) be seeded with srand. If we don't want that, we'd have to use something like [EigenRand](https://bab2min.github.io/eigenrand/v0.3.0/en/getting_started.html). But I agree with @sjahr's notion, that he is replacing existing functionality with a simpler implementation.

include/geometric_shapes/random_number_utils.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+  /// \brief Return a random quaternion. The reason for this wrapper function is that we need to make sure, that srand()
+  /// is seeded before UnitRandom() is called. The seeding takes place in the Singleton's constructor.
+  [[nodiscard]] auto getRandomQuaternion(std::optional<unsigned int> seed = std::nullopt)
+  {
+    if (seed.has_value())
+    {
+      // Optional re-seed
+      std::srand(seed.value());
+    }
```
rhaschke, line 133: Why would you want to re-seed in calls to `getRandomQuaternion()`?

include/geometric_shapes/bodies.h
```diff
@@ -225,7 +225,7 @@ class Body
      Sometimes multiple attempts need to be generated.
      The function terminates with failure (returns false) after \e max_attempts attempts.
      If the call is successful (returns true) the point is written to \e result */
-  virtual bool samplePointInside(random_numbers::RandomNumberGenerator& rng, unsigned int max_attempts,
+  virtual bool samplePointInside(std::function<double(double, double)> random_value, unsigned int max_attempts,
```
rhaschke, line 228: Readability is reduced here: What's the meaning of the arguments and return value of the std::function?
If this approach is to be followed, there should be a type definition for this function type explaining the args.

rhaschke (commented):
Sorry, but I don't yet get the reading behind this PR. IMO this PR achieves the following:
- It replaces boost types with std types, which I definitely support.
- It moves the RNG definition from the RNG package into this package. Why?
  This reduces modularity.
- It introduces _thread-local_ RNG _singletons_. Why?
  One motivation could be found [here](https://stackoverflow.com/questions/8813592/c11-thread-safety-of-random-number-generators). But, I would like to see the argumentation for this both in the PR comment and a code comment :wink:
- It changes public API all over. Is this really needed?

I suggest to stick to the concept of a `RandomNumberGenerator` implemented in a separate package. Whether this should be a singleton or thread-local, is the user's decision isn't it?

include/geometric_shapes/random_number_utils.hpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+      thread_local RandomNumberGenerator instance(seed_sequence.value());
+      return instance;
+    }
+    else
+    {
+      thread_local RandomNumberGenerator instance;
+      return instance;
+    }
+  }
```
tylerjw:
```suggestion
  /** The first time this function is called it creates a thread_local random number generator.
    * If a seed sequence is provided on that first call it is used to create the generator, otherwise the random device is used to seed the generator.
    * After the first call to this function, if a seed sequence is provided this function throws.
    */
  [[nodiscard]] static RandomNumberGenerator& getInstance(std::optional<std::seed_seq> seed_sequence = std::nullopt)
  {
    bool first = false;
    thread_local RandomNumberGenerator instance = [&seed_sequence, &first]() {
      first = true;
      if (seed_sequence.has_value()) {
        return RandomNumberGenerator{seed_sequence};
      } else {
        return RandomNumberGenerator{};
      }
    }();
    if (!first && seed_sequence.has_value()) {
      throw std::exception{"RandomNumberGenerator cannot be re-seeded"};
    }
    return instance;
  }
```

include/geometric_shapes/random_number_utils.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+public:
+  /** The first time this function is called it creates a thread_local random number generator.
+   * If a seed sequence is provided on that first call it is used to create the generator, otherwise the random device
+   * is used to seed the generator. After the first call to this function, if a seed sequence is provided this function
+   * throws.
+   */
+  [[nodiscard]] static RandomNumberGenerator& getInstance(std::optional<std::seed_seq> seed_sequence = std::nullopt)
+  {
+    bool first = false;
```
tylerjw: first is per-thread so it should also be `thread_local`

```suggestion
    thread_local bool first = false;
```

include/geometric_shapes/random_number_utils.hpp
```diff
@@ -0,0 +1,123 @@
...
+    thread_local RandomNumberGenerator instance = [&seed_sequence, &first]() {
```
tylerjw: You don't need to capture the thread_local `first` variable because it is static (assuming you accept my previous change to make it thread_local).

```suggestion
    thread_local RandomNumberGenerator instance = [&seed_sequence]() {
```

include/geometric_shapes/random_number_utils.hpp
sjahr: Initially, srand is seeded with NULL but for testing or other cases where a deterministic random number is required I've added this option

include/geometric_shapes/bodies.h
sjahr: Good point, I've updated the comment

sjahr:
> Sorry, but I don't yet get the reading behind this PR. IMO this PR achieves the following:
> 
> * It replaces boost types with std types, which I definitely support.
> * It moves the RNG definition from the RNG package into this package. Why?
>   This reduces modularity.
> * It introduces _thread-local_ RNG _singletons_. Why?
>   One motivation could be found [here](https://stackoverflow.com/questions/8813592/c11-thread-safety-of-random-number-generators). But, I would like to see the argumentation for this both in the PR comment and a code comment wink
> * It changes public API all over. Is this really needed?
> 
> I suggest to stick to the concept of a `RandomNumberGenerator` implemented in a separate package. Whether this should be a singleton or thread-local, is the user's decision isn't it?

@rhaschke Thank you for this review and please excuse my long response time. The motivation of this PR was to remove the dependency on the random numbers package in order to reduce the number of packages to maintain and to make use of `random`. This might be at the cost of modularity but I think in this case the advantages outweight and the change of the interface to
```
  virtual bool samplePointInside(std::function<double(double, double)> random_value, unsigned int max_attempts,
 ```
 allows the users to use their own random number generation algorithms.
 
As you pointed out, the reason for the thread-local singleton is on one hand thread-safety and in addition to that, I wanted to make the usage computationally more efficient by motivation the user to minimize the number of constructions of random number generator instances.

I've added more comments to the class definition to document these motivations

peci1:
I also don't understand the motivation behind this PR.

Removing a dependency by copying its code inside the dependent package is not the way modular systems are built. Moreover, random_numbers is a simple package that only depends on boost.

Apart from the architectural point of view, I also don't have much understanding for the mathematical meaning. You suggest that with this PR users can provide their custom RNG implementations. However, all uses of the RNG in this package expect an RNG that draws from a strictly bounded distribution. I know only one such distribution - the uniform one. Okay, in an extreme case, you could use a "sloped uniform" distribution, but that's weird - do you think if the user wants the distribution to have some slope along the X axis of a box, that he will want to have the same slope along the other two axes? Okay, so why not passing 1-3 lambdas to the samplePointInside functions, one for each axis? I don't like the direction where it's heading. If, on the other hand, you'd pass an unbounded distribution (like Gaussian), the implicit contract of the sampling methods would have been broken and the sampling might take extremely long because there could be much more misses of the shape's volume than expected. I assume if someone needs to sample Gaussians from a box, then it would probably deserve a more clever algorithm than the generic one from this library. Or do you have some concrete example of a non-uniform distribution you'd like to pass to the sampling methods?

tylerjw:
The boost random generator was created before there was a good implementation of random number generation in the stl. In some ways what is in the stl now resembles the interface in boost in that it is annoyingly complex to use correctly but it does provide industrial strength randomness.

We hear the complaint of copying code into this package going in the wrong direction. We also have the constraint that updating random_numbers itself to be more modern means dealing with all the code that depends on the interfaces of random_numbers that we would break. It seems easier to us to replace random_numbers with a new package we can incrementally adopt. We are working on doing that work here: https://github.com/PickNikRobotics/RSL/blob/main/src/rand.cpp

Once this is ready for use (released) we will adapt this PR to use this new package that contains a random number generator function.

sjahr:
> I also don't understand the motivation behind this PR.
> 
> Removing a dependency by copying its code inside the dependent package is not the way modular systems are built. Moreover, random_numbers is a simple package that only depends on boost.
> 
> Apart from the architectural point of view, I also don't have much understanding for the mathematical meaning. You suggest that with this PR users can provide their custom RNG implementations. However, all uses of the RNG in this package expect an RNG that draws from a strictly bounded distribution. I know only one such distribution - the uniform one. Okay, in an extreme case, you could use a "sloped uniform" distribution, but that's weird - do you think if the user wants the distribution to have some slope along the X axis of a box, that he will want to have the same slope along the other two axes? Okay, so why not passing 1-3 lambdas to the samplePointInside functions, one for each axis? I don't like the direction where it's heading. If, on the other hand, you'd pass an unbounded distribution (like Gaussian), the implicit contract of the sampling methods would have been broken and the sampling might take extremely long because there could be much more misses of the shape's volume than expected. I assume if someone needs to sample Gaussians from a box, then it would probably deserve a more clever algorithm than the generic one from this library. Or do you have some concrete example of a non-uniform distribution you'd like to pass to the sampling methods?

As Tyler said, due to the feedback we got, I reconsidered and we added a minimal package with the proposed random number generation instead of reducing the whole package to a simple header.

Regrading your second statement I am a bit confused because a using a different random number generator does not necessarily imply that a different distribution is used. The [random](https://en.cppreference.com/w/cpp/numeric/random) package contains multiple random number engines (algorithms to create random numbers) that can be used to create a bounded uniform distribution. In addition to that, it might be useful for testing to create a function that creates specific values and to pass it as fake random number generator. From my understanding, this change increases modularity because users don't have to use our generator implementation based on the mersenne twister algorithm. Please correct me if I am missing something here

peci1:
> From my understanding, this change increases modularity because users don't have to use our generator implementation based on the mersenne twister algorithm. Please correct me if I am missing something here

You're right that the generator and distribution are two distinct parts. I just did not see too much value in allowing use of other generators (well, you want a uniform distribution in the end, so you should not generate too much differently than MT). If you'd provide a deterministic fake generator, these sampling functions could even get broken. The only real difference could be usage of a true RNG, but that seems like an overkill to me for this use-case...

tylerjw (commented):
Now that RSL is released and synced on Humble and Rolling, what do you think of changing this PR to depend on the random header from RSL instead of random numbers?

The advantage of using the one from RSL is it depends on the much more modern random number generation from the STL instead of from boost.

Here is the header for random from RSL: https://github.com/PickNikRobotics/RSL/blob/main/include/rsl/random.hpp

sjahr:
Closing this since I don't plan on picking this up anytime soon

PR #225, opened 2022-08-23
getScaledDimensions(): avoid vtable lookup

rhaschke:
As the concrete type is known anyway, we can skip the vtable lookup.

### DIFF
--- a/src/body_operations.cpp
+++ b/src/body_operations.cpp
@@ -86,19 +86,19 @@ shapes::ShapeConstPtr bodies::constructShapeFromBody(const bodies::Body* body)
   {
     case shapes::SPHERE:
     {
-      const auto& dims = static_cast<const bodies::Sphere*>(body)->getScaledDimensions();
+      const auto& dims = static_cast<const bodies::Sphere*>(body)->bodies::Sphere::getScaledDimensions();
       result.reset(new shapes::Sphere(dims[0]));
       break;
     }
     case shapes::BOX:
     {
-      const auto& dims = static_cast<const bodies::Box*>(body)->getScaledDimensions();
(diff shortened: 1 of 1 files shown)


peci1:
I wasn't sure whether the compiler optimizations do not skip the vtable lookup anyways, and it seems there is actually some difference even with O2 optimizations: https://godbolt.org/z/fEE49WEdT.

LGTM.

peci1 (approved):

src/body_operations.cpp
```diff
@@ -86,19 +86,19 @@ shapes::ShapeConstPtr bodies::constructShapeFromBody(const bodies::Body* body)
   {
     case shapes::SPHERE:
     {
-      const auto& dims = static_cast<const bodies::Sphere*>(body)->getScaledDimensions();
+      const auto& dims = static_cast<const bodies::Sphere*>(body)->bodies::Sphere::getScaledDimensions();
```
peci1, line 90: Maybe I'd add a comment explaining this unusual syntax.

```suggestion
      // The syntax used for calling getScaledDimensions() allows skipping virtual dispatch - we already know the exact type of the body.
      const auto& dims = static_cast<const bodies::Sphere*>(body)->bodies::Sphere::getScaledDimensions();
```

rhaschke: I added a slightly different comment.

simonschmeisser:
> I wasn't sure whether the compiler optimizations do not skip the vtable lookup anyways, and it seems there is actually some difference even with O2 optimizations: https://godbolt.org/z/fEE49WEdT.
> 

It should be able to do this optimization if the function is marked `final` and ~most likely~ also if the derived 'leaf' class is marked `final`

https://godbolt.org/z/48dnYzE78

PR #228, opened 2023-01-19
Switch to clang-format-14

henningkayser:
https://github.com/ros-planning/moveit2/issues/1736


henningkayser:
Well, look at that. CI is failing because the old clang-format-10 version is not available in Jammy...

rhaschke (approved):

PR #232, opened 2023-04-06
Improve accuracy of computed oriented bounding boxes

peci1:
In case one OBB is fully contained inside the other, merging them together will now result in the larger one being returned exactly (normally, FCL yields a larger OBB than the original one).

This change is needed to keep multiple merged OBBs coming from a robot model sane. I don't know exactly why the original PR adding OBBs support did not contain this improvement - my code in robot_body_filter has been doing this for the last 4 years: https://github.com/peci1/robot_body_filter/blob/2c13fc8b0aa5ae7ad71a336a3916f53171c8d9f5/src/utils/obb.cpp#L122-L140 .

With this change, our tracked robot model OBB is kept within +- 2 meters range. Without this change, after being composed by merging 40+ OBBs with various rotations, the overall OBB dimensions are in the range of 40 kilometers! My understanding is that [FCL first computes an average orientation of both OBBs](https://github.com/flexible-collision-library/fcl/blob/54e9619bc2b084ee50e986ac3308160d663481c4/src/BV/OBB.cpp#L128) (altough a bit weird by averaging coordinates of the quaternions) and then finds the smallest box with this orientation. This way, after merging, the resulting box may be up to `sqrt(3)`-times larger. Thus a model composed of 40 OBBs can have the resulting OBB up to `sqrt(3)^40` larger than necessary. This PR stops this infinite growth quite early once an OBB large enough to contain the whole model is created.

I added a few tests that fail when fully relying on FCL to merge the two OBBs.

### DIFF
--- a/src/obb.cpp
+++ b/src/obb.cpp
@@ -42,23 +42,30 @@ public:
 
 OBB::OBB()
 {
-  this->obb_.reset(new OBBPrivate);
+  obb_.reset(new OBBPrivate);
+  // Initialize the OBB to position 0, with 0 extents and identity rotation
 #if FCL_MAJOR_VERSION > 0 || FCL_MINOR_VERSION > 5
-  // Zero-initialize the OBB. This is done to unify the behavior of FCL 0.5 (which zero-initializes by default) and
-  // FCL 0.6, which leaves the fields uninitialized.
-  this->obb_->extent.setZero();
-  this->obb_->To.setZero();
-  this->obb_->axis.setZero();
+  // FCL 0.6+ does not zero-initialize the OBB.
+  obb_->extent.setZero();
+  obb_->To.setZero();
+  obb_->axis.setIdentity();
+#else
+  // FCL 0.5 zero-initializes the OBB, so we just put the identity into orientation.
+  obb_->axis[0][0] = 1.0;
+  obb_->axis[1][1] = 1.0;
+  obb_->axis[2][2] = 1.0;
 #endif
 }
 
-OBB::OBB(const OBB& other) : OBB()
+OBB::OBB(const OBB& other)
 {
+  obb_.reset(new OBBPrivate);
   *obb_ = *other.obb_;
 }
 
-OBB::OBB(const Eigen::Isometry3d& pose, const Eigen::Vector3d& extents) : OBB()
+OBB::OBB(const Eigen::Isometry3d& pose, const Eigen::Vector3d& extents)
 {
+  obb_.reset(new OBBPrivate);
   setPoseAndExtents(pose, extents);
 }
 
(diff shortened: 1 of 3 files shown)


src/obb.cpp
```diff
@@ -43,6 +43,13 @@ class OBBPrivate : public FCL_OBB
 OBB::OBB()
 {
   this->obb_.reset(new OBBPrivate);
+#if FCL_MAJOR_VERSION > 0 || FCL_MINOR_VERSION > 5
+  // Zero-initialize the OBB. This is done to unify the behavior of FCL 0.5 (which zero-initializes by default) and
```
rhaschke: Didn't you recently add code to explicitly initialize the OBB at the caller side (https://github.com/ros-planning/geometric_shapes/pull/231)?
This should be reverted then, shouldn't it?

rhaschke (approved):
Approving with a small nitpick.

src/obb.cpp
peci1: No, these two complement each other. I've tested #231 on Melodic where it worked as expected. However, on Noetic with FCL 0.6, the assumed zero-initialization does not happen. This PR adds just the zero-initialization to FCL 0.6 so that the behavior is the same for both.

It is helpful to note that zero-initialization of the OBB is different from "identity"-initialization. The FCL OBB structure contains a 3x3 matrix for rotation. If zero-initialized, it is invalid.

#231 added the "identity"-initialization so that the merging of OBBs does not start with an invalid one.

peci1: Another option would be to always "identity"-initialize the OBB in the default constructor.

The zero-initialized OBBs were just a direct translation from FCL 0.5 where zero-initialization is the default. However, it seems to be a very bug-prone approach as invalid OBBs might quite easily get into the code.

I suggest a default-constructed OBB should be a valid one with identity rotation, zero pose and zero extents. "Uninitialized"/empty OBBs would then be represented as all OBBs with zero extents (but with valid orientation).

If this change would be made, #231 could get reverted.

Do you like this approach?

rhaschke: I think that's the right way to go, particularly if you want to avoid invalid OBBs.

peci1: Ok, I'll update the PR.

peci1: Done. I also added a check for the "identity"-initialization by no-arg constructor.

peci1:
Dear maintainers, merging this PR would help resolving build regression in https://github.com/peci1/robot_body_filter/issues/23 which currently blocks Noetic sync: https://discourse.ros.org/t/preparing-for-noetic-sync-2023-04-13/30848 . Would you please have a look at it shortly so that we can move on? Thanks! @AndyZe you helped reviewing the [original PR adding OBB support](https://github.com/ros-planning/geometric_shapes/pull/105)

src/obb.cpp
```diff
@@ -42,16 +42,30 @@ class OBBPrivate : public FCL_OBB
...
+  obb_->axis[1][1] = 1.0;
+  obb_->axis[2][2] = 1.0;
+#endif
 }
 
-OBB::OBB(const OBB& other) : OBB()
+OBB::OBB(const OBB& other)
 {
+  obb_.reset(new OBBPrivate);
   *obb_ = *other.obb_;
```
rhaschke: What about using the copy constructor of OBBPrivate?
```suggestion
  obb_.reset(new OBBPrivate(*other.obb_);
```

src/obb.cpp
```diff
@@ -94,20 +108,12 @@ void OBB::getPose(Eigen::Isometry3d& pose) const
...
 #if FCL_MAJOR_VERSION == 0 && FCL_MINOR_VERSION == 5
-  if (!obb_->axis[0].isZero() && !obb_->axis[3].isZero() && !obb_->axis[2].isZero())
-  {
-    pose.linear().col(0) = fromFcl(obb_->axis[0]);
-    pose.linear().col(1) = fromFcl(obb_->axis[1]);
-    pose.linear().col(2) = fromFcl(obb_->axis[2]);
-  }
+  pose.linear().col(0) = fromFcl(obb_->axis[0]);
+  pose.linear().col(1) = fromFcl(obb_->axis[1]);
+  pose.linear().col(2) = fromFcl(obb_->axis[2]);
```
rhaschke, line 113: This is much cleaner now. :+1:

rhaschke (approved):
Consider the suggested improvement of copy constructor.

src/obb.cpp
peci1: Thanks for the suggestion. Applied.

src/obb.cpp
peci1: Yes, I also like it ;)

peci1:
Thanks for merging, @rhaschke . Could you please do another noetic release with this fix, so that we can use it in our package?

And also please consider a melodic backport and release (a simple cherry pick should do). Thanks.

rhaschke:
Can you please prepare the backport? Thanks.

peci1:
Sure! #233.

rhaschke:
https://github.com/ros/rosdistro/pull/36804

PR #233, opened 2023-04-12
Improve accuracy of computed oriented bounding boxes [melodic backport of #232]

peci1:
Backport of #232.

### DIFF
--- a/include/geometric_shapes/obb.h
+++ b/include/geometric_shapes/obb.h
@@ -53,8 +53,7 @@ class OBBPrivate;
 class OBB
 {
 public:
-  /** \brief Initialize an oriented bounding box at position 0, with 0 extents and
-   *         identity orientation. */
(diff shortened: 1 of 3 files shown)


include/geometric_shapes/obb.h
```diff
@@ -52,6 +53,8 @@ class OBBPrivate;
 class OBB
 {
 public:
+  /** \brief Initialize an oriented bounding box at position 0, with 0 extents and
+   *         identity orientation. */
```
rhaschke:
```suggestion
  /** Initialize an oriented bounding box at position 0, with 0 extents and identity orientation. */
```

rhaschke (commented):
Minor formatting improvement - didn't notice that before.

include/geometric_shapes/obb.h
peci1: Thanks, suggestion added.

PR #235, opened 2023-09-29
[ROS-O] drop c++11 requirement

v4hn:
modern gtest breaks without c++14.

Looks like we missed one until now. :)


simonschmeisser (approved):
I'm a bit confused why that never was a problem

simonschmeisser:
To resolve my confusion: this package does not use ros logging and therefore did not have the "default" c++17 vs 11 or 14 issue

rhaschke (approved):

Issue #127, opened 2020-03-25
Wrong assumption about planes of a ConvexMesh in case padding is used


peci1:
In `ConvexMesh::useDimensions()`, there is this code that processes planes (halfspaces) generated by qhull:

https://github.com/ros-planning/geometric_shapes/blob/07a5254b698ecc89e511dc5b3182ecb6c7d2258e/src/bodies.cpp#L903-L904

It checks if the plane for the currently processed triangle/facet isn't too similar to the plane for the previous triangle. If it is, it merges these planes into one and makes both triangles reference the same plane.

However, according to the definition of padding for convex meshes in this library, this optimization goes wrong when non-zero padding is used.

I made some visualizations to make this argument clearly visible:

![scale_line](https://user-images.githubusercontent.com/182533/77491853-39352280-6e3f-11ea-9f92-5a6acbf11fa1.png)

Here, the red line is original (unscaled) "mesh" (well...). The cyan one is the line after applying only scaling. And the green one is after padding is applied. You can see that the line stops being a line and any planes that would be optimized by the previous algorithm would no longer be parallel.

![scale_box](https://user-images.githubusercontent.com/182533/77491855-3b977c80-6e3f-11ea-88a7-0e3950de5c55.png)

Maybe even more obvious.

This has also consequences for more code paths. It seems to me that two things are required.

1. Get rid of the "if" in the above code excerpt and just add one plane per each triangle. 
1. Take a broader look around the code, and e.g. add `mesh_data_->scaled_planes_` which would contain the correctly scaled and padded planes which could be used for collision checking. For example this code tries to subtract padding from the data, but fails to do it correctly (a part of it is fixed in #109):

https://github.com/ros-planning/geometric_shapes/blob/07a5254b698ecc89e511dc5b3182ecb6c7d2258e/src/bodies.cpp#L1091-L1100

If there were an already precomputed field containing the scaled planes, all computations of this kind could be done both efficiently and without fear if everything is correctly accounted for. My approach to recomputing the scaled planes would be to just take the scaled vertices and reconstruct the scaled planes from them. It has just one drawback - so far, `mesh_data_` are not touched in `updateInternalData()`, however if we wanted to keep a set of scaled planes, these would obviously need to be updated in `updateInternalData()`. But this function anyways has a loop over all vertices to recompute the scaled vertices, so I think another loop for scaled planes is not a problem. It might well be compensated by the possible speedups from #126.

v4hn:
[...]

PR #238, opened 2024-04-02
Pad using weighted vertex normals

### LINKED ISSUE #76: Non-uniform padding of non-convex, long meshes
I noticed that when applying a padding factor to non-convex, long-in-one dimension meshes, such as the table pictured below, the padding applied to the vertices is not uniform, since it is applied along the vector joining the mesh center and the vertex being padded. Any advice on how to make the padding more accurate for this kind of meshes?

Test case example: I have a robot arm on top of the table depicted below and want to add padding of 10 cm to the table to avoid the robot hand to collide during pick and place, meaning that I'd expect to have the table's top, for instance, "grow up" 10 cm (besides "growing in the X and Y directions as well).

In reality, when I apply padding to the table, what I get is that the table top grows in X ~9cm, whereas in Y and Z it grows only ~3cm.

table

kbrameld:
Resolves: #76

As discussed in the linked issue, padding behavior for complex meshes are not good. These changes improve upon the padding functionality, and obtain a closer functionality to blender's fatten method. Two main changes:
* padding is performed in direction of vertex normal instead of direction from center of mesh
* calculates better weighted vertex normal by accounting for the angle of a face at a vertex.

A complex STL file (original) | Current padding method
:----:|:----:
Screenshot from 2024-04-02 15-01-38 |  Screenshot from 2024-04-02 14-01-02

Fatten method (in Blender) | New padding method
:----:|:----:
Screenshot from 2024-04-02 11-39-09 | Screenshot from 2024-04-02 14-27-05

Non-uniform padding x-direction | y-direction | z-direction
:----:|:----:|:----:
Screenshot from 2024-04-02 15-00-00 |  Screenshot from 2024-04-02 15-32-50 | Screenshot from 2024-04-02 15-32-27

As discussed in the linked issue, although this is a general improvement, this change may potentially affect behavior in downstream packages if they relied on the bad padding.

### DIFF
--- a/src/shapes.cpp
+++ b/src/shapes.cpp
@@ -564,6 +564,70 @@ void Mesh::computeVertexNormals()
   }
 }
 
+void Mesh::computeWeightedVertexNormals()
+{
+  if (!triangle_normals)
+    computeTriangleNormals();
+  if (vertex_count && !vertex_normals)
+    vertex_normals = new double[vertex_count * 3];
+  EigenSTL::vector_Vector3d avg_normals(vertex_count, Eigen::Vector3d(0, 0, 0));
+
+  for (unsigned int tIdx = 0; tIdx < triangle_count; ++tIdx)
+  {
+    unsigned int tIdx3 = 3 * tIdx;
+    unsigned int tIdx3_1 = tIdx3 + 1;
+    unsigned int tIdx3_2 = tIdx3 + 2;
+
+    unsigned int v1 = triangles[tIdx3];
+    unsigned int v2 = triangles[tIdx3_1];
+    unsigned int v3 = triangles[tIdx3_2];
+
+    // Get angles for each vertex at this triangle
+    Eigen::Vector3d p1{ vertices[3 * v1], vertices[3 * v1 + 1], vertices[3 * v1 + 2] };
+    Eigen::Vector3d p2{ vertices[3 * v2], vertices[3 * v2 + 1], vertices[3 * v2 + 2] };
+    Eigen::Vector3d p3{ vertices[3 * v3], vertices[3 * v3 + 1], vertices[3 * v3 + 2] };
+
+    {
+      // Use re-arranged dot product equation to calculate angle between two vectors
+      auto angleBetweenVectors = [](const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) -> double {
+        return std::acos(v1.dot(v2) / (v1.norm() * v2.norm()));
+      };
+
+      // Use law of cosines to compute angles
+      auto ang1 = angleBetweenVectors(p2 - p1, p3 - p1);
+      auto ang2 = angleBetweenVectors(p1 - p2, p3 - p2);
+      auto ang3 = angleBetweenVectors(p1 - p3, p2 - p3);
+
+      // Weight normal with angle
+      avg_normals[v1][0] += triangle_normals[tIdx3] * ang1;
+      avg_normals[v1][1] += triangle_normals[tIdx3_1] * ang1;
+      avg_normals[v1][2] += triangle_normals[tIdx3_2] * ang1;
+
+      avg_normals[v2][0] += triangle_normals[tIdx3] * ang2;
+      avg_normals[v2][1] += triangle_normals[tIdx3_1] * ang2;
+      avg_normals[v2][2] += triangle_normals[tIdx3_2] * ang2;
+
+      avg_normals[v3][0] += triangle_normals[tIdx3] * ang3;
+      avg_normals[v3][1] += triangle_normals[tIdx3_1] * ang3;
+      avg_normals[v3][2] += triangle_normals[tIdx3_2] * ang3;
+    }
+  }
+
+  for (std::size_t i = 0; i < avg_normals.size(); ++i)
+  {
+    auto& avg_normal = avg_normals[i];
+    if (avg_normal.squaredNorm() != 0.0)
+    {
+      avg_normal.normalize();
+    }
+
+    unsigned int i3 = i * 3;
+    vertex_normals[i3] = avg_normal[0];
+    vertex_normals[i3 + 1] = avg_normal[1];
+    vertex_normals[i3 + 2] = avg_normal[2];
+  }
+}
+
 void Mesh::mergeVertices(double threshold)
 {
   const double thresholdSQR = threshold * threshold;
@@ -623,7 +687,7 @@ void Mesh::mergeVertices(double threshold)
   if (triangle_normals)
     computeTriangleNormals();
   if (vertex_normals)
-    computeVertexNormals();
+    computeWeightedVertexNormals();
 }
 
 } /* namespace shapes */
--- a/src/shapes.cpp
+++ b/src/shapes.cpp
@@ -407,8 +407,8 @@ void Mesh::scaleAndPadd(double scaleX, double scaleY, double scaleZ, double padd
 
     // Padding in each direction
     vertices[i3] = scaledX + vertex_normals[i3] * paddX;
-    vertices[i3+1] = scaledY + vertex_normals[i3+1] * paddY;
-    vertices[i3+2] = scaledZ + vertex_normals[i3+2] * paddZ;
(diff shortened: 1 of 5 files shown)


peci1:
Wow, it's been a long time since I was interested in this issue.

In #127, I found:

> Another comment was that generally change of the current behavior is okay as long as we can provide some guarantees on the relationship between the old approach and the new one (i.e. all points contained by the old behavior would be contained by the new one, or so...).

Could you sketch up a "proof" of some relationship?

Or, alternatively, wouldn't it be easier to create a function with a new name (`fatten`?) instead of `scaleAndPad` to retain the original behavior for existing code?

kbrameld:
> Another comment was that generally change of the current behavior is okay as long as we can provide some guarantees on the relationship between the old approach and the new one (i.e. all points contained by the old behavior would be contained by the new one, or so...).

Not sure if there is a nice relationship. The old padding method was closer to a scaling operation, and the resulting mesh often didn't include the original vertices. Here's an example with the blue being the padded shapes:

Old method | New Method
:----:|:----:
Screenshot from 2024-03-14 14-54-19 | Screenshot from 2024-03-14 14-46-05

The only meshes that will look similar between the old and new method will be meshes where the weighted vertex normal direction is close to the direction of the vertex from the center. (eg. cube, sphere meshes) For any other mesh, the padded shapes will look different.

> Or, alternatively, wouldn't it be easier to create a function with a new name (fatten?) instead of scaleAndPad to retain the original behavior for existing code?

Yes, this can be considered an alternative. If we go this approach, I advocate for deprecating the "pad" methods (in upcoming ros2 distros).

src/shapes.cpp
```diff
@@ -564,6 +556,70 @@ void Mesh::computeVertexNormals()
...
+    unsigned int tIdx3_2 = tIdx3 + 2;
+
+    unsigned int v1 = triangles[tIdx3];
+    unsigned int v2 = triangles[tIdx3_1];
+    unsigned int v3 = triangles[tIdx3_2];
+
+    // Get angles for each vertex at this triangle
+    Eigen::Vector3d p1{ vertices[3 * v1], vertices[3 * v1 + 1], vertices[3 * v1 + 2] };
+    Eigen::Vector3d p2{ vertices[3 * v2], vertices[3 * v2 + 1], vertices[3 * v2 + 2] };
+    Eigen::Vector3d p3{ vertices[3 * v3], vertices[3 * v3 + 1], vertices[3 * v3 + 2] };
```
rhaschke: Consider using `Eigen::Map<Eigen::Vector3d>(vertices + 3*vi, 3)` to avoid copying the data.

src/shapes.cpp
```diff
@@ -564,6 +556,70 @@ void Mesh::computeVertexNormals()
   }
 }
 
+void Mesh::computeWeightedVertexNormals()
+{
+  if (!triangle_normals)
+    computeTriangleNormals();
+  if (vertex_count && !vertex_normals)
+    vertex_normals = new double[vertex_count * 3];
+  EigenSTL::vector_Vector3d avg_normals(vertex_count, Eigen::Vector3d(0, 0, 0));
```
rhaschke: Consider using an `Eigen::Map<Eigen::MatrixXd>` to `vertex_normals` (a vertex_count x 3 matrix) to avoid allocating an extra array (and final data copying).

src/shapes.cpp
```diff
@@ -564,6 +556,70 @@ void Mesh::computeVertexNormals()
   }
 }
 
+void Mesh::computeWeightedVertexNormals()
```
rhaschke: As far as I see, this method replaces the old method `computeVertexNormals()`.

src/shapes.cpp
```diff
@@ -564,6 +556,70 @@ void Mesh::computeVertexNormals()
...
+
+    // Get angles for each vertex at this triangle
+    Eigen::Vector3d p1{ vertices[3 * v1], vertices[3 * v1 + 1], vertices[3 * v1 + 2] };
+    Eigen::Vector3d p2{ vertices[3 * v2], vertices[3 * v2 + 1], vertices[3 * v2 + 2] };
+    Eigen::Vector3d p3{ vertices[3 * v3], vertices[3 * v3 + 1], vertices[3 * v3 + 2] };
+
+    {
+      // Use re-arranged dot product equation to calculate angle between two vectors
+      auto angleBetweenVectors = [](const Eigen::Vector3d& v1, const Eigen::Vector3d& v2) -> double {
+        return std::acos(v1.dot(v2) / (v1.norm() * v2.norm()));
```
rhaschke: Don't you need to handle the special case |v1| = 0 or |v2| = 0, i.e. dividing zero by zero?

rhaschke (commented):
Thanks for this contribution.
Given that the old mesh padding was not working as expected, I don't see the need to introduce the new behavior with a new name. I even suggest replacing the old method `computeVertexNormals()` with the new one.

peci1:
> Given that the old mesh padding was not working as expected, I don't see the need to introduce the new behavior with a new name.

At least robot_body_filter users actively use padding with its current behavior. Any big changes could break their workflows. But as long as the result is similar for the commonly found meshes (e.g. UR-5), it should be okay. I think the example mesh used in this PR is kind of extreme and very far from what people normally use, so the big difference in the example is okay.

peci1:
@kbrameld Do I get it correctly that the unexpected case of enlarging a square would still be present with this PR? https://github.com/ros-planning/geometric_shapes/issues/127#issue-587387573 .

src/shapes.cpp
kbrameld: I like this suggestion. To match data storage order, either we need a 3xN or Nx3 (with Row major). Not sure if there is much difference to the two, so I'm going to go with ``Eigen::Map<Eigen::Matrix<double, 3, Eigen::Dynamic>>``.

src/shapes.cpp
kbrameld: Changed to replace current method rather than adding a new method

src/shapes.cpp
kbrameld: Good catch, updated

kbrameld:
> @kbrameld Do I get it correctly that the unexpected case of enlarging a square would still be present with this PR? [#127 (comment)](https://github.com/ros-planning/geometric_shapes/issues/127#issue-587387573) .

You're correct, the changes in this PR won't change the behavior of enlarging the square in that issue.

src/shapes.cpp
```diff
@@ -575,48 +576,49 @@ void Mesh::computeWeightedVertexNormals()
...
+      return std::acos(vec1.dot(vec2) / (vec1_norm * vec2_norm));
+    };
+
+    // Use law of cosines to compute angles
+    auto ang1 = angleBetweenVectors(p2 - p1, p3 - p1);
+    auto ang2 = angleBetweenVectors(p1 - p2, p3 - p2);
+    auto ang3 = angleBetweenVectors(p1 - p3, p2 - p3);
+
+    // Weight normal with angle
+    mapped_normals.col(v1)[0] += triangle_normals[tIdx3] * ang1;
```
rhaschke: If you Eigen::Map `triangle_normals` as well, this code further simplifies to vector addition.

kbrameld: Sounds good, addressed

rhaschke (approved):

ijnek:
Once this is merged, I will create a PR against the ``ros2`` branch too. What's required to move forward with this PR?

ijnek:
Pinging this thread for attention to be reviewed

rhaschke:
Merging this, w/o a second review as nobody seems to do it...

kbrameld:
Thanks for reviewing @rhaschke!

peci1:
I've tested robot_body_filter (which uses geometric_shapes) with a few models from https://github.com/ctu-vras/rosdevday_cloud_filtering and the models seem to be affected only a little by this PR. So it seems this change should be safe for most users! Thanks.

timonegk:
> Once this is merged, I will create a PR against the `ros2` branch too.

Now that this is merged, could you create a PR for ROS 2?

PR #239, opened 2024-04-19
[ROS2] Porting bodies::Body::computeBoundingBox changes from noetic to ROS2

spelletier1996:
This is the ROS2 port of the changes from #210

Needed to eventually port [robot_body_filter](https://github.com/peci1/robot_body_filter/pull/27) to ROS2

We initially tried to git cherry-pick directly but ran into issues hence the extra commits. 

Adds fcl to the build, the obb files, computeBoundingBox (obb version), mergeBoundingBoxesApprox, and related boundingbox tests


include/geometric_shapes/bodies.h
```diff
@@ -240,6 +241,10 @@ class Body
       pose. Scaling and padding are accounted for. */
   virtual void computeBoundingBox(AABB& bbox) const = 0;
 
+  /** \brief Compute the oriented bounding box for the body, in its current
+     pose. Scaling and padding are accounted for. */
+  virtual void computeBoundingBox(OBB& bbox) const = 0;
```
peci1, line 246: This will break ABI (the virtual keyword). Not sure what is the current policy regarding ABI breakages.

include/geometric_shapes/obb.h
```diff
@@ -0,0 +1,142 @@
+// Copyright 2019 Open Robotics
```
peci1:
```suggestion
// Copyright 2024 Open Robotics
```

include/geometric_shapes/body_operations.h
```diff
@@ -25,7 +25,6 @@
 // CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
-
```
peci1: Unnecesary whitespace change.

package.xml
```diff
@@ -36,6 +36,15 @@
...
   <build_depend>assimp-dev</build_depend>
+  <!-- <build_depend>boost</build_depend> -->
+  <!-- <build_depend>eigen</build_depend>
+  <build_depend>eigen_stl_containers</build_depend> -->
+  <!-- <build_depend>libconsole-bridge-dev</build_depend> -->
+  <!--
+  <build_depend condition="$ROS_DISTRO == noetic">fcl</build_depend> -->
+  <!-- <build_depend>libqhull</build_depend>
+  <build_depend>octomap</build_depend> -->
+  <build_depend>libfcl-dev</build_depend>
```
peci1: This apparently needs more care.

src/aabb.cpp
```diff
@@ -30,14 +30,19 @@
 
 #include <geometric_shapes/aabb.h>
 
+#include <fcl/config.h>
+#if FCL_MAJOR_VERSION > 0 || FCL_MINOR_VERSION > 5
+#include <fcl/geometry/shape/utility.h>
+#endif
+
```
peci1, line 34: FCL 0.5 has already disappeared from all relevant platforms. All ROS 2 Jazzy platorms from REP 2000 currently offer FCL 0.7 . So all these checks for FCL 0.5 should be removed.

src/aabb.cpp
```diff
@@ -30,14 +30,19 @@
...
 void bodies::AABB::extendWithTransformedBox(const Eigen::Isometry3d& transform, const Eigen::Vector3d& box)
 {
+#if FCL_MAJOR_VERSION == 0 && FCL_MINOR_VERSION == 5
```
peci1: Remove the FCL 0.5 check.

src/obb.cpp
```diff
@@ -0,0 +1,206 @@
+#include <geometric_shapes/obb.h>
+
+#include <fcl/config.h>
+
+#if FCL_MAJOR_VERSION == 0 && FCL_MINOR_VERSION == 5
```
peci1: Remove FCL 0.5 check.

src/obb.cpp
```diff
@@ -0,0 +1,206 @@
...
+typedef fcl::OBB FCL_OBB;
+#else
+#include <fcl/math/bv/OBB.h>
+typedef fcl::Vector3d FCL_Vec3;
+typedef fcl::OBB<double> FCL_OBB;
+#endif
+
+namespace bodies
+{
+#if FCL_MAJOR_VERSION == 0 && FCL_MINOR_VERSION == 5
```
peci1: Remove the FCL 0.5 checks for both `toFcl` and `fromFcl`. Or, even better, remove `toFcl` and `fromFcl` at all as they're no longer needed.

src/obb.cpp
```diff
@@ -0,0 +1,206 @@
...
+{
+public:
+  using FCL_OBB::OBB;
+};
+
+OBB::OBB()
+{
+  obb_.reset(new OBBPrivate);
+  // Initialize the OBB to position 0, with 0 extents and identity rotation
+#if FCL_MAJOR_VERSION > 0 || FCL_MINOR_VERSION > 5
```
peci1: Remove FCL 0.5 check.

src/obb.cpp
```diff
@@ -0,0 +1,206 @@
...
+{
+  *obb_ = *other.obb_;
+  return *this;
+}
+
+void OBB::setPoseAndExtents(const Eigen::Isometry3d& pose, const Eigen::Vector3d& extents)
+{
+  const auto rotation = pose.linear();
+
+#if FCL_MAJOR_VERSION == 0 && FCL_MINOR_VERSION == 5
```
peci1: Remove FCL 0.5 check.

src/obb.cpp
```diff
@@ -0,0 +1,206 @@
...
+  Eigen::Vector3d extents;
+  getExtents(extents);
+  return extents;
+}
+
+void OBB::getPose(Eigen::Isometry3d& pose) const
+{
+  pose = Eigen::Isometry3d::Identity();
+  pose.translation() = fromFcl(obb_->To);
+#if FCL_MAJOR_VERSION == 0 && FCL_MINOR_VERSION == 5
```
peci1: Remove FCL 0.5 check,

peci1 (changes requested):
The PR looks good to me in general.

It needs a few changes I left as review comments. Most importantly, getting rid of the code complications needed to support FCL 0.5 .

Note for maintainers: This PR contains the original #210 PR adding OBBs as well as all the later fixups required to make it work.

Also, this PR breaks ABI. I'm not sure if that is a problem or not. If it is a problem, we can implement the same temporary workarounds that were done in https://github.com/ros-planning/geometric_shapes/commit/b00b1e54e6a6d133d3bb8b634161b1715051c829 .

src/obb.cpp
spelletier1996: To confirm, the FCL 0.5 checks are to be removed, the code inside the if check is kept, and the code inside the else block is removed for all 0.5 checks?

spelletier1996:
I've removed to and from FCL and the related 0.5 checks

src/aabb.cpp
```diff
@@ -53,10 +50,4 @@ void bodies::AABB::extendWithTransformedBox(const Eigen::Isometry3d& transform,
   const Eigen::Vector3d v_delta(x_range, y_range, z_range);
   extend(t + v_delta);
   extend(t - v_delta);
-#else
-  fcl::AABBd aabb;
-  fcl::computeBV(fcl::Boxd(box), transform, aabb);
-  extend(aabb.min_);
-  extend(aabb.max_);
-#endif
```
peci1: I think you removed the wrong ifdef branch. The one with computeBV should be retained.

src/obb.cpp
```diff
@@ -78,22 +46,14 @@ void OBB::setPoseAndExtents(const Eigen::Isometry3d& pose, const Eigen::Vector3d
...
 
-#if FCL_MAJOR_VERSION == 0 && FCL_MINOR_VERSION == 5
-  obb_->axis[0] = toFcl(rotation.col(0));
-  obb_->axis[1] = toFcl(rotation.col(1));
-  obb_->axis[2] = toFcl(rotation.col(2));
-#else
   obb_->axis = rotation;
-#endif
-
-  obb_->To = toFcl(pose.translation());
```
peci1: Shouldn't you keep `obb_->To = pose.translation();` here?

CMakeLists.txt
```diff
@@ -66,6 +67,7 @@ set(THIS_PACKAGE_EXPORT_DEPENDS
   resource_retriever
   shape_msgs
   visualization_msgs
+  fcl
```
peci1: I don't think fcl needs to be exported as a dependency. It is only used from .cpp files and not from the .h files. So it should only be build and export dependency, not build_export dependency.

src/aabb.cpp
```diff
@@ -30,14 +30,16 @@
 
 #include <geometric_shapes/aabb.h>
 
+#include <fcl/config.h>
+#include <fcl/geometry/shape/utility.h>
+
```
peci1, line 34: These includes now seem irrelevant.

package.xml
```diff
@@ -36,6 +36,8 @@
   <depend>visualization_msgs</depend>
 
   <build_depend>assimp-dev</build_depend>
+  <build_depend>eigen</build_depend>
+  <build_depend>libfcl-dev</build_depend>
```
peci1, line 40: `<exec_depend>libfcl-dev</exec_depend>` is also needed.

peci1 (changes requested):
Thanks for the changes. I went through one more round of reviews and you'll find them again as code comments.

peci1:
@rhaschke @henningkayser Is support for Foxy and Galactic still required? If so, their underlying Ubuntu 20.04 still distributes FCL 0.5, which would mean this PR would need to retain all the FCL 0.5 checks.

henningkayser:
> @rhaschke @henningkayser Is support for Foxy and Galactic still required? If so, their underlying Ubuntu 20.04 still distributes FCL 0.5, which would mean this PR would need to retain all the FCL 0.5 checks.

we just merged support for Jazzy and dropped Foxy and Galactic. Could you update this branch to rerun CI @spelletier1996?

src/aabb.cpp
spelletier1996: With switching to the correct ifdef as you mentioned below I believe that #include <fcl/geometry/shape/utility.h>
is still required?

CMakeLists.txt
spelletier1996: Im still newer to cmake, in that case would I just remove it from this export set and the rest can remain the same?

package.xml
tmayoff: Having both the `<build_depend>libfcl-dev</build_depend>` and `<exec_depend>libfcl-dev</exec_depend>` caused the CI to fail, is it enough to have it in either exec_depend or just `<depend>...</depend>`

peci1: This is wrong. `<exec_depend>libfcl-dev` has to be there.

src/aabb.cpp
peci1: Right, some fcl include is still required. This seems good.

peci1:
@spelletier1996 Please, do the one update of package.xml I asked. With that done, I hope this PR would be ready for merging.

package.xml
peci1: Or, even better, `<exec_depend>libfcl</exec_depend>` (but wait until https://github.com/ros/rosdistro/pull/43530  is merged).

tmayoff:
I'm not sure how available @spelletier1996 is to make changes, so I went ahead and modified the `package.xml` file to re-add the libfcl-dev

If the linked rosdistro PR for libfcl gets merged let me know and I can edit this again to remove the -dev

rhaschke (approved):
Looks good to me from high-level perspective. Relying on @peci1's detailed review.

peci1 (approved):

peci1:
@tmayoff https://github.com/ros/rosdistro/pull/43530 just landed, adding key `libfcl` to reference the runtime of the system version of FCL library. You can now change `<exec_depend>` to use `libfcl`.

PR #252, opened 2024-11-26
Forward port ABI compatibility workaround with OBBs to ros2 branch

sea-bass:
Forward porting https://github.com/moveit/geometric_shapes/commit/b00b1e5 since we ran into the ABI incompatibility...

https://github.com/moveit/moveit2/issues/3130

Closes https://github.com/moveit/geometric_shapes/issues/251

### DIFF
--- a/include/geometric_shapes/bodies.h
+++ b/include/geometric_shapes/bodies.h
@@ -242,8 +242,10 @@ public:
   virtual void computeBoundingBox(AABB& bbox) const = 0;
 
   /** \brief Compute the oriented bounding box for the body, in its current
-     pose. Scaling and padding are accounted for. */
-  virtual void computeBoundingBox(OBB& bbox) const = 0;
+   * pose. Scaling and padding are accounted for.
+   * \note This function should be pure virtual, but it can't in order to maintain ABI compatibility.
+   **/
+  void computeBoundingBox(OBB& bbox) const;
 
   /** \brief Get a clone of this body, but one that is located at the pose \e pose */
   inline BodyPtr cloneAt(const Eigen::Isometry3d& pose) const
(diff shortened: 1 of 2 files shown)


sjahr (approved):
Thanks!

rhaschke:
Just applying the ABI compatibility hack is a bad idea. As pointed out by @peci1 in its original ROS1 commits, this should only be necessary for compatibility with release binaries.
However, at least in CI you are building everything (geometric_shapes and MoveIt) from source. Thus, there shouldn't be any ABI compat issues. As we see issues anyway, they probably originate from the build system not correctly resolving headers. I suggest to tackle this root cause instead of introducing the ABI compat hack here.

rhaschke:
I will have a short look this afternoon.

sea-bass:
Thank you! This issue was specifically happening with the `main` branch of `moveit2` built from source in a Ubuntu 22.04/Humble environment, if that is any help.

(To verify changes, I was just running the `humble-ci` job)

rhaschke:
I strongly believe that this the good old header installation problem. Since ROS Humble, headers should be installed to `/opt/ros/<distro>/include/<package>/<package>/`:
https://docs.ros.org/en/humble/Releases/Release-Humble-Hawksbill.html#c-headers-are-installed-in-a-subdirectory
https://colcon.readthedocs.io/en/released/user/overriding-packages.html#install-headers-to-a-unique-include-directory

MoveIt2 never realized that change, causing ABI issues with package overlays to persist.
We recently fixed this for srdfdom: https://github.com/moveit/srdfdom/pull/132 and https://github.com/moveit/srdfdom/commit/65ce909735f95d781e22496672bd9a8e35fc6baf.
I think, it's time to go through all MoveIt-releated packages and eventually fix it. 
Unfortunately, I don't have the resources to do so.

rhaschke:
I created https://github.com/moveit/moveit2/issues/3134 as a reminder and will provide a PR here as another example.

sea-bass:
MoveIt has had some active new contributors lately that may want to/be able to help with this!

Do you mind making a new issue on that repo? I can bring it up at the maintainer meeting alongside the topic of switching from .h to .hpp files.

Edit: hah, way ahead of me :)

PR #253, opened 2024-11-27
Install headers into subdirectory

rhaschke:
- Reverts the ABI compatibility hack
- Implements https://github.com/moveit/geometric_shapes/pull/252#issuecomment-2503816583


sea-bass (approved):
Thanks!

I can prep a next release for this package and we'll do the temporary uninstall of binaries for the time being on `moveit2`.

PR #256, opened 2024-12-05
Do not export fcl dependency

### LINKED ISSUE #255: Likely regression with dependencies in 2.3.0
#239 adds a `<build_depend>` on `libfcl-dev` and `<exec_depend>` on `libfcl` - however when 2.3.0 was released, `robot_calibration` debians stopped building for Jazzy and Rolling with the following error:

```
/usr/bin/ld: cannot find -lfcl: No such file or directory
```

I don't directly do anything with fcl, but do find_package on geometric_shapes and link against (to utilize the mesh loader).

`geometric_shapes` has only three other (released) downstream users: `moveit_core`, `moveit_ros_occupancy_map_monitor`, `moveit_ros_visualization` - the latter two of which depend on `moveit_core`. `moveit_core` has a `<depend>` on `libfcl-dev`. Thus, these continued to build properly.

I've temporarily patched my build by adding a `<depend>libfcl-dev</depend>` to the package.xml of robot_calibration - but not sure if this maybe indicates a deeper problem that could catch other users of `geometric_shapes` off guard.

peci1:
Hopefully closes #255.

It seems we have erroneousely exported `fcl` as interface link library when it should only be private (it is only used in `obb.cpp` and in no header files).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,7 @@ target_compile_options(${PROJECT_NAME} PRIVATE ${PROJECT_COMPILE_OPTIONS})
 target_include_directories(${PROJECT_NAME}
     PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
            $<INSTALL_INTERFACE:include/${PROJECT_NAME}>)
-target_link_libraries(${PROJECT_NAME} fcl)
+target_link_libraries(${PROJECT_NAME} PRIVATE fcl)
 ament_target_dependencies(${PROJECT_NAME}
   ${THIS_PACKAGE_EXPORT_DEPENDS}
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,15 +90,12 @@ target_compile_options(${PROJECT_NAME} PRIVATE ${PROJECT_COMPILE_OPTIONS})
 target_include_directories(${PROJECT_NAME}
     PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
            $<INSTALL_INTERFACE:include/${PROJECT_NAME}>)
-target_link_libraries(${PROJECT_NAME} PRIVATE fcl)
 ament_target_dependencies(${PROJECT_NAME} PUBLIC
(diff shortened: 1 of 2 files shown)


rhaschke (approved):

peci1:
Hmm, it seems this PR has started a bit of a rabbit hole.

To fix the first build failure, I've added the `PUBLIC` keyword to the first `ament_target_dependencies`.

I've missed the second `ament_target_dependencies`, which should also receive a `PUBLIC` keyword, because that is the only one `ament_target_dependencies` accepts (it doesn't support `INTERFACE` or `PRIVATE`).

Now, I could add the `PUBLIC` keyword there, but assimp is not an exported dependency according to package.xml (the same way as fcl). On the other hand, qhull is declared as `<depends>`, so it should be exported.

However, the only usage of qhull I see is in bodies.cpp, so it could technically also be a non-transitive dependency.

Please, maintainers, guide me towards the wanted result in these decisions:

1. Should I add `PUBLIC` keyword to the `ament_target_dependencies(... assimp QHULL)` ?
2. Should I move `assimp` out of the `ament_target_dependencies` and put it next to `fcl` in `target_link_libraries(PRIVATE)` ? This would remove `assimp` library from the exported link libraries, probably breaking some downstream that depends on it (which would be wrong because assimp is not transitive according to package.xml).
3. Or should I move both `assimp` and `QHULL` to the `target_link_libraries(PRIVATE)` and redeclare `qhull` as non-transitive dependency in package.xml? This would remove both assimp and qhull from exported link libraries, probably breaking some downstream that depends on them (which would be wrong in case of assimp, but correct in case of qhull, except the fact that relying on transitive dependencies is not a good habit).

rhaschke:
Dependencies that are publicly exposed in headers should be declared public in cmake and transitive in package.xml.
As far as I see, neither assimp, fcl, or qhull are exposed in the headers. Thus, you should go for option 3 and declare all of them private and non-transitive.
Issues in downstream packages should be fixed later on.

peci1:
Okay, I've turned fcl, assimp and qhull into private dependencies. Let's see what happens downstream :)

peci1:
Feel free to squash-merge this as the commit history in this PR does not make much sense.

rhaschke (approved):

PR #261, opened 2025-05-30
fix missing lib

mosfet80:
Compile failed : missing Boost::random


rhaschke:
Correctly fixed via #262 and https://github.com/moveit/random_numbers/commit/e16611ab8359b0e3fd9288bad760c4189df29c49

PR #266, opened 2026-01-14
Fixed CMakeLists.txt: Boost random

ahcorde:
Related to https://build.ros2.org/view/Rdev/job/Rdev__geometric_shapes__ubuntu_noble_amd64/18/console#console-section-3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ find_package(Eigen3 REQUIRED)
 
 find_package(ament_cmake REQUIRED)
 find_package(assimp REQUIRED)
+find_package(Boost REQUIRED random)
 find_package(console_bridge REQUIRED)
 find_package(console_bridge_vendor REQUIRED)
 find_package(eigen_stl_containers REQUIRED)


rhaschke (commented):
While this is a quick fix, `geometric_shapes` doesn't directly depend on Boost::random:
```
$ grep -r "boost/"
test/test_ray_intersection.cpp:#include <boost/filesystem.hpp>
test/test_loaded_meshes.cpp:#include <boost/filesystem.hpp>
test/test_bounding_sphere.cpp:#include <boost/filesystem.hpp>
test/test_point_inclusion.cpp:#include <boost/filesystem.hpp>
test/test_body_operations.cpp:#include <boost/filesystem.hpp>
include/geometric_shapes/shape_messages.h:#include <boost/variant.hpp>
src/mesh_operations.cpp:#include <boost/math/constants/constants.hpp>
src/bodies.cpp:#include <boost/math/constants/constants.hpp>
```
Rather, `random_numbers` should pull in that dependency in a transitive fashion.
I'm surprised that this is not yet working as `random_numbers` [correctly declares the dependency](https://github.com/moveit/random_numbers/blob/b74f37a2456da5e978b33746167934923a78c50a/ConfigExtras.cmake#L31):
```cmake
find_package(Boost REQUIRED COMPONENTS random)
```

rhaschke:
> I'm surprised that this is not yet working as `random_numbers` [correctly declares the dependency](https://github.com/moveit/random_numbers/blob/b74f37a2456da5e978b33746167934923a78c50a/ConfigExtras.cmake#L31).

Ah. The reason is that the referenced buildfarm build uses an outdated version of `random_numbers` (2.0.2 instead of 2.0.4):
```
$ apt policy ros-rolling-random-numbers
ros-rolling-random-numbers:
  Installed: (none)
  Candidate: 2.0.4-1noble.20250730.134617
  Version table:
     2.0.4-1noble.20250730.134617 500
        500 http://packages.ros.org/ros2/ubuntu noble/main amd64 Packages
```
Hence, you should check why the buildfarm uses an outdated version.

rhaschke:
Looks like you found the issue - the [build succeeded now](https://build.ros2.org/job/Rdev__geometric_shapes__ubuntu_noble_amd64/19/consoleFull).
What was the reason for the outdated `random_numbers` package?

ahcorde:
The package was not rebuild with the new `random_number` release

PR #272, opened 2026-07-19
Include cassert in check_isometry.h

Tobias-Fischer:
This is part of an effort to contribute RoboStack downstream patches back upstream.

This upstreams `patch/ros-rolling-geometric-shapes.patch`, authored in RoboStack by Silvio Traversaro `<silvio@traversaro.it>`.

`check_isometry.h` defines `ASSERT_ISOMETRY` with `assert`, so it should include `<cassert>` directly where the macro is defined instead of adding the include to an unrelated translation unit.

### DIFF
--- a/include/geometric_shapes/check_isometry.h
+++ b/include/geometric_shapes/check_isometry.h
@@ -32,7 +32,7 @@
 #define GEOMETRIC_SHAPES_CHECK_ISOMETRY_H
 
 #include <algorithm>
-#include <assert.h>
+#include <cassert>
 #include <iostream>
 #include <math.h>
 #include <Eigen/Core>


Tobias-Fischer:
FYI @traversaro, this is a small PR from the effort to upstream RoboStack downstream patches tracked in https://github.com/RoboStack/robostack.github.io/issues/16.

rhaschke (commented):
I don't see an assert used in this particular cpp file.

Tobias-Fischer:
Good catch - I think that should now be the correct header @rhaschke

traversaro:
> FYI @traversaro, this is a small PR from the effort to upstream RoboStack downstream patches tracked in [RoboStack/robostack.github.io#16](https://github.com/RoboStack/robostack.github.io/issues/16).

Thanks!

rhaschke (commented):
cassert eventually includes assert.h.
Where do you see the benefit?

Tobias-Fischer:
Thanks, you are right. I re-checked the full header set: `check_isometry.h` is the only public header using `assert` / `ASSERT_ISOMETRY`, and upstream already includes `<assert.h>` there. So this PR no longer fixes a missing include; it is just a stylistic `<assert.h>` to `<cassert>` change.

Sorry for the noise - I will close this as unnecessary.

Tobias-Fischer:
Closing as unnecessary after re-checking: the relevant public header already includes assert.h, so this PR is only a stylistic cassert change.

